PR #1, opened 2020-10-15
Update some of the .idea project configuration files

Sigmonia:
#### Rationale
Update some of the .idea files for the new Git world

#### Changes
Change the projects VCS type to Git
Update the root .gitignore file to ignore most of the .idea folder

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,12 @@ server/testAutomation/
 
 # Include the modules build.gradle file -- this line will override previous exclusion
 !server/modules/build.gradle
+
+# Ignore parts of the .idea dir
+.idea/libraries
+.idea/modules
+.idea/.gitignore
+.idea/jarRepositories.xml
+.idea/modules.xml
+.idea/server.iml
+.idea/workspace.xml
--- a/.gitignore
+++ b/.gitignore
@@ -21,8 +21,8 @@ server/testAutomation/
 !server/modules/build.gradle
 
 # Ignore parts of the .idea dir
-.idea/libraries
-.idea/modules
+.idea/libraries/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -19,3 +19,12 @@ server/testAutomation/
 
 # Include the modules build.gradle file -- this line will override previous exclusion
 !server/modules/build.gradle
+
+# Ignore parts of the .idea dir
+.idea/libraries
+.idea/modules
+.idea/.gitignore
```
labkey-susanh, line 26: This is presumably generated by intelliJ.  Interesting that it doesn't think the libraries and modules directories are to be ignored, but whatever.

.gitignore
```diff
@@ -19,3 +19,12 @@ server/testAutomation/
...
+.idea/jarRepositories.xml
+.idea/modules.xml
+.idea/server.iml
```
labkey-susanh: I don't have this file.

.gitignore
```diff
@@ -19,3 +19,12 @@ server/testAutomation/
 
 # Include the modules build.gradle file -- this line will override previous exclusion
 !server/modules/build.gradle
+
+# Ignore parts of the .idea dir
+.idea/libraries
+.idea/modules
```
labkey-susanh: Do these two directories need trailing slashes?

Sigmonia:
```suggestion
.idea/libraries/
.idea/modules/
```

.gitignore
Sigmonia:
```suggestion
```

labkey-susanh (approved):

PR #15, opened 2020-12-01
bump LabKey version

labkey-klum:
#### Rationale
As part of the monthly release cycle we will bump the LabKey version on develop to 21.01-SNAPSHOT

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -37,7 +37,7 @@ buildFromSource=true
 
 # The default version for LabKey artifacts that are built or that we depend on.
 # override in an individual module's gradle.properties file as necessary 
-labkeyVersion=21.01-SNAPSHOT
+labkeyVersion=21.1-SNAPSHOT
 labkeyClientApiVersion=1.3.2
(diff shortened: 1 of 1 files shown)


labkey-tchad (commented):
I though the leading zeros were just for the issue tracker milestone. I think our GitHub release workflows will get confused by this and I suspect other scripts might also.

gradle.properties
```diff
@@ -37,7 +37,7 @@ buildFromSource=true
 
 # The default version for LabKey artifacts that are built or that we depend on.
 # override in an individual module's gradle.properties file as necessary 
-labkeyVersion=20.12-SNAPSHOT
+labkeyVersion=21.01-SNAPSHOT
```
labkey-tchad:
```suggestion
labkeyVersion=21.1-SNAPSHOT
```

labkey-tchad (changes requested):

gradle.properties
labkey-klum: good catch!

labkey-tchad (approved):

PR #18, opened 2020-12-03
Use Gradle Plugin DSL for applying plugins and specifying versions

labkey-susanh:
#### Rationale
We're updating how we publish our plugins to move away from the older method and streamlining how the plugins get applied so we can avoid some duplication.

#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/105

#### Changes
* add pluginManagement section to settings.gradle file
* use plugin DSL for applying various
* move application of versioning plugin into base plugin

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,4 +1,34 @@
 
+pluginManagement {
+    repositories {
+        maven {
+            url "${artifactory_contextUrl}/plugins-release"
+        }
+        if (gradlePluginsVersion.contains("SNAPSHOT") || versioningPluginVersion.contains("SNAPSHOT")) {
+            maven {
+                url "${artifactory_contextUrl}/plugins-snapshot-local"
+            }
+        }
+    }
+    plugins {
+        id 'org.labkey.build.antlr' version "${gradlePluginsVersion}" apply false
+        id 'org.labkey.build.api' version "${gradlePluginsVersion}" apply false
+        id 'org.labkey.build.base' version "${gradlePluginsVersion}" apply false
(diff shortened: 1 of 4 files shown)


settings.gradle
```diff
@@ -1,4 +1,34 @@
...
+        id 'org.labkey.build.jsdoc' version "${gradlePluginsVersion}" apply false
+        id 'org.labkey.build.module' version "${gradlePluginsVersion}" apply false
+        id 'org.labkey.build.multiGit' version "${gradlePluginsVersion}" apply false
+        id 'org.labkey.build.npmRun' version "${gradlePluginsVersion}" apply false
+        id 'org.labkey.build.serverDeploy' version "${gradlePluginsVersion}" apply false
+        id 'org.labkey.build.teamCity' version "${gradlePluginsVersion}" apply false
+        id 'org.labkey.build.testRunner' version "${gradlePluginsVersion}" apply false
+        id 'org.labkey.build.tomcat' version "${gradlePluginsVersion}" apply false
+        id 'org.labkey.build.xsddoc' version "${gradlePluginsVersion}" apply false
+    }
```
labkey-tchad, line 30: Do we need to specify the versioning plugin version?
```suggestion
        id 'org.labkey.versioning' version "${versioningPluginVersion}" apply false
    }
```

labkey-susanh: Yes, I wasn't sure that was going to work since we publish that plugin differently, but it seems to do the right thing.  I can apply this plugin to an individual module using the plugins closure in its build.gradle file. I'm not sure if we'll ever have need for that, but nice to have common treatment.

labkey-tchad (approved):

PR #21, opened 2020-12-18
Item 8021 : Introduction of Spring boot with Embedded Tomcat

labkey-ankurj:
#### Rationale
This work provides the infrastructure required to run LabKey server on embedded tomcat. This utilizes SpringBoot which starts the project from server/embeddded/LabKeyServer which is the main class for the configuration. 

#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/108
* https://github.com/LabKey/platform/pull/1853

#### Changes
* new server/embedded module
* new gradle property `useEmbeddedTomcat` to switch between regular deployment vs embedded deployment


server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -0,0 +1,358 @@
...
+
+				if (!webAppLocationPresent)
+				{
+					var destDirectory = new File("").getAbsolutePath() + "/server";
+					webAppLocation = destDirectory + "/labkeywebapp";
+					boolean extracted = new File(webAppLocation).exists();
+
+					if (!extracted)
+					{
+						try (JarFile jar = new JarFile("labkeyServer-21.1-SNAPSHOT.jar"))
```
labkey-ankurj: Should I iterate here to in the current path to find the jar file?

gradle.properties
```diff
@@ -84,7 +84,7 @@ activationVersion=1.2.1
 
 annotationsVersion=15.0
 
-apacheTomcatVersion=8.5.51
+apacheTomcatVersion=9.0.34
```
labkey-ankurj: @labkey-susanh mentioned to check if we can move forward with the change to tomcat version here.

labkey-adam: Won't this make us incompatible with Tomcat 8.5.x? We haven't discussed dropping support for that version.

labkey-ankurj: I can try to change the version of embedded tomcat.

labkey-adam: Yes, embedding 9.0.x would be great. But we should continue to compile LabKey against 8.5.x.

gradle.properties
```diff
@@ -257,3 +257,6 @@ xercesImplVersion=2.11.0
 xmlApisVersion=1.0.b2
 
 xmlbeansVersion=3.0.1
+
+# uncomment the following line for running the application with embedded tomcat
+useEmbeddedTomcat
```
labkey-jeckels: Is this intended to be commented out?

server/configs/application.properties
```diff
@@ -0,0 +1,15 @@
+server.port=8081
+
+labkey.dataSourceName[0]=jdbc/labkeyDataSource
```
labkey-jeckels: Let's refine these property names. I propose a prefix like "labkeyDatabase" or "labkeyDataSource" instead of "labkey". Otherwise, it's hard to tell what you're supposed to be configuring from just the .properties file.

server/embedded/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+    main {
+        java {
+            srcDirs = ['src']
+        }
+    }
+}
+
+dependencies {
+    implementation 'org.springframework.boot:spring-boot-starter-web:2.4.0'
+    implementation 'org.apache.tomcat.embed:tomcat-embed-core:9.0.34'
```
labkey-jeckels: This is what's defining which Tomcat is being used for embedding. It should be OK to split this from the top-level 8.5 version number.

But it seems useful to have a property for the Tomcat version we're using. Right now it's a mix of 9.0.34 and 9.0.37 which doesn't seem correct

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -0,0 +1,358 @@
...
+import java.util.zip.ZipInputStream;
+
+@SpringBootApplication
+public class LabKeyServer
+{
+	private static final int BUFFER_SIZE = 4096;
+
+	public static void main(String[] args)
+	{
+		ApplicationContext ctx = SpringApplication.run(LabKeyServer.class, args);
```
labkey-jeckels: Indentation looks like it's 8-spaces here instead of our normal 4.

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -0,0 +1,358 @@
...
+
+				// for development, point to the local deploy/labkeyWebapp directory
+				boolean webAppLocationPresent = props.getWebAppLocation() != null;
+				var webAppLocation = "";
+
+				if (!webAppLocationPresent)
+				{
+					var destDirectory = new File("").getAbsolutePath() + "/server";
+					webAppLocation = destDirectory + "/labkeywebapp";
+					boolean extracted = new File(webAppLocation).exists();
```
labkey-jeckels: Let's talk this through. It should be OK to have multiple JARs for different versions in the same directory, and be able to start with any of them and have it do the right thing.

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -0,0 +1,358 @@
...
+				{
+					ContextResource dataSourceResource = new ContextResource();
+					dataSourceResource.setName(props.getDataSourceName().get(i));
+					dataSourceResource.setAuth("Container");
+					dataSourceResource.setType(DataSource.class.getName());
+					dataSourceResource.setProperty("driverClassName", props.getDriverClassName().get(i));
+					dataSourceResource.setProperty("url", props.getUrl().get(i));
+					dataSourceResource.setProperty("password", props.getPassword().get(i));
+					dataSourceResource.setProperty("username", props.getUsername().get(i));
+					dataSourceResource.setProperty("maxTotal", "20");
```
labkey-jeckels: All of these properties should be supported via the .properties file too. OK to defer them to the next story.

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -0,0 +1,358 @@
...
+
+			private ContextResource getMailResource()
+			{
+				// Get session/mail properties
+				MailProperties mailProps = smtpSource();
+				ContextResource mailResource = new ContextResource();
+				mailResource.setName("mail/Session");
+				mailResource.setAuth("Container");
+				mailResource.setType("javax.mail.Session");
+				mailResource.setProperty("mail.smtp.host", mailProps.smtpHost);
```
labkey-jeckels: We'll need more properties here too. At least the ones captured here:

https://www.labkey.org/Documentation/wiki-page.view?name=cpasxml#SMTPsettings

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -0,0 +1,358 @@
...
+				mailResource.setProperty("mail.smtp.port", mailProps.smtpPort);
+				return mailResource;
+			}
+		};
+	}
+
+	/**
+	 * Extracts a zip file specified by the zipFilePath to a directory specified by
+	 * destDirectory (will be created if does not exists)
+	 * @param zipInputStream
```
labkey-jeckels: Useless JavaDoc - delete these three lines or add info about the params and exception

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -0,0 +1,358 @@
...
+
+				for (int i = 0; i < numOfDataResources; i++)
+				{
+					ContextResource dataSourceResource = new ContextResource();
+					dataSourceResource.setName(props.getDataSourceName().get(i));
+					dataSourceResource.setAuth("Container");
+					dataSourceResource.setType(DataSource.class.getName());
+					dataSourceResource.setProperty("driverClassName", props.getDriverClassName().get(i));
+					dataSourceResource.setProperty("url", props.getUrl().get(i));
+					dataSourceResource.setProperty("password", props.getPassword().get(i));
```
labkey-jeckels: This will blow up with an ugly stack trace if you have more URLs than you have passwords, etc.

labkey-jeckels (changes requested):

server/embedded/src/org/labkey/embedded/LabKeyServer.java
labkey-ankurj: looks correct locally.

server/configs/application.properties
```diff
@@ -1,15 +1,33 @@
...
+server.port=8443
+server.ssl.enabled=true
+server.ssl.enabled-protocols=TLSv1.3,TLSv1.2,TLSv1.1
+server.ssl.key-alias=my_selfsigned
+server.ssl.key-store=/path/to/key-store-file
+server.ssl.key-password=pwd
+server.ssl.key-store-type=key-store-type
+server.ssl.ciphers=ciphers
+
+labkeyDataSource.dataSourceName[0]=jdbc/labkeyDataSource
```
labkey-jeckels: @labkey-adam Do you have ideas on these property names? I can't decide which I think is best.

I like qualifying the name itself with "labkey" but it's a bit confusing to have all of them include "labkeyDataSource" when the value also needs to be "jdbc/labkeyDataSource"

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -167,22 +164,94 @@ private ContextResource getMailResource()
...
-	 * @param zipInputStream
-	 * @param destDirectory
-	 * @throws IOException
-	 */
-	public static void extractZip(InputStream zipInputStream, String destDirectory) throws IOException
+	private static void extractExecutableJar(String destDirectory, String jarFilePath)
+	{
+		try
+		{
+			// check for multiple jars and blow up if multiple are present
```
labkey-jeckels: The check is happening elsewhere, so we should move the comment.

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -167,22 +164,94 @@ private ContextResource getMailResource()
...
+			throw new ConfigException("Executable jar not found.");
+		}
+
+		// only 1 executable jar should be there
+		if (jarCount == 1)
+		{
+			return jarFileName;
+		}
+
+		throw new ConfigException("Multiple executable jars found. Must provide only one executable jar");
```
labkey-jeckels: Why not build a Set<String> so we can give an even better error message?

And since we're not checking if the JAR is actually executable, I'd just say "jar"

server/embedded/src/org/labkey/embedded/LabKeyServer.java
labkey-jeckels: Indentation still seems wrong. Is it saved with tabs instead of spaces?

labkey-ankurj: interesting.. it doesn't have tabs and looks fine locally.

server/embedded/build.gradle
labkey-adam: Why not embed the most recent Tomcat 9.0.x release? That would be 9.0.41.

server/configs/application.properties
labkey-adam: And I'm assuming one can configure other data sources, which wouldn't be named `jdbc/labkeyDataSource`. So you'd have:

```labkeyDataSource.dataSourceName[1]=jdbc/someOtherDataSource```

Which is really confusing.

It looks like other labkey.xml properties (webAppLocation, masterEncryptionKey) are prefaced with `context`. Seems reasonable to be consistent with all properties from the context file, so:

```context.dataSourceName[1]=jdbc/someOtherDataSource```

File: server/embedded/src/org/labkey/embedded/LabKeyServer.java (hunk 6 above)
Comment (labkey-jeckels) [reply server/embedded/src/org/labkey/embedded/LabKeyServer.java]: I just pulled down the source and it's using tabs. Try doing View->Active Editor->Show Whitespace when you've got one of these new embedded files open. Also, check the bottom right where it's showing Tab.

image

Comment (labkey-ankurj) [reply server/embedded/src/org/labkey/embedded/LabKeyServer.java]: thanks, corrected.

File: server/embedded/src/org/labkey/embedded/LabKeyServer.java [hunk 15]
```diff
@@ -0,0 +1,472 @@
...
+                MailProperties mailProps = smtpSource();
+                ContextResource mailResource = new ContextResource();
+                mailResource.setName("mail/Session");
+                mailResource.setAuth("Container");
+                mailResource.setType("javax.mail.Session");
+                mailResource.setProperty("mail.smtp.host", mailProps.getSmtpHost());
+                mailResource.setProperty("mail.smtp.user", mailProps.getSmtpUser());
+                mailResource.setProperty("mail.smtp.port", mailProps.getSmtpPort());
+
+                if (mailProps.getSmtpFrom() != null &&
```
labkey-jeckels: I don't think these should be all-or-nothing. Setting a from address shouldn't mean that you have to have a passowrd or custom socket factory, or vice versa.

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -0,0 +1,472 @@
...
+    }
+
+    private static String getExecutableJar(String currentPath) throws ConfigException
+    {
+        File currentDir = new File(currentPath);
+        List<String> jarsPresent = new ArrayList<>();
+
+        for (File file: currentDir.listFiles())
+        {
+            if (file.getName().endsWith(".jar"))
```
labkey-jeckels: Do a .toLowerCase() here?

labkey-jeckels (changes requested):

server/embedded/src/org/labkey/embedded/LabKeyServer.java
labkey-ankurj: sure

server/embedded/src/org/labkey/embedded/LabKeyServer.java
labkey-ankurj: ok, changed

server/configs/application.properties
```diff
@@ -0,0 +1,33 @@
+# server.port=8080
+
+## To switch to http, uncomment above server config and comment
+## below server config
+
+server.port=8443
+server.ssl.enabled=true
+server.ssl.enabled-protocols=TLSv1.3,TLSv1.2,TLSv1.1
+server.ssl.key-alias=my_selfsigned
+server.ssl.key-store=/path/to/key-store-file
```
labkey-jeckels: This is fine for now but we'll want to think how we want devs to set these properties. Right now they'll have to edit both mssql.properties/pg.properties and application.properties, and be sure not to check in any of them.

labkey-jeckels (approved):

.gitignore
```diff
@@ -4,6 +4,7 @@
 build/
 enlistment.properties
 server/config.properties
+server/embedded/config/application.properties
```
labkey-susanh: This is no longer needed here.  We now put this file in the build directory.

PR #28, opened 2021-01-15
Issue42107: Module Import Permit Mixed-Case

RosalineP:
#### Rationale
Upload of new module archive rejects with a "Module name doesn't match" message if a .module file's name is not undercase-only, to match with the exploded directory name.

#### Related Pull Requests
* n/a

#### Changes
* Permit .module upload if file name matches exploded directory name irrespective of casing.


RosalineP:
Note: Manual test successful.

labkey-matthewb (approved):

PR #56, opened 2021-02-22
Add server/modules/README.md

labkey-tchad:
#### Rationale
Git doesn't allow empty directories so `server/modules` won't exist in a fresh enlistment. Adding a placeholder, similar to the one in `remoteapi/README.md`

#### Changes
* Add 'server/modules/README.md'

### DIFF
--- a/server/modules/README.md
+++ b/server/modules/README.md
@@ -1,0 +1,2 @@
+Place modules' source in this directory. \
+See [LabKey documentation](https://www.labkey.org/Documentation/wiki-page.view?name=devMachine) for more information on configuring your development environment.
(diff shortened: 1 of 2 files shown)


server/modules/README.md
```diff
@@ -0,0 +1,4 @@
+### Module Checkout Directory
+
+Clone module source into this directory. Gradle will look for modules here when building. \
```
labkey-susanh, line 3: Is the backslash at the end there intended?

labkey-susanh (approved):

server/modules/README.md
labkey-tchad: Yeah. It adds a line break in markdown.

PR #57, opened 2021-02-24
Issue 40038: Ride-or-die Mode - default to shutting down on failures

labkey-jeckels:
#### Rationale
Embedded Tomcat is intended to participate more fully in container-based deployments, which want the server to shut down if it hits a fatal error

#### Changes
* Set a system property if it's not explicitly set to tell the server to shut down the VM if it hits a fatal startup problem


labkey-ankurj (approved):
Where is this prop used?

labkey-ankurj:
nvm just saw the related PR

PR #66, opened 2021-04-06
Add templates for some IntelliJ config files, a task to install these, and an updated gitIgnore file

labkey-susanh:
#### Rationale
Most of the files in the `.idea` directory are updated by IntelliJ as a developer does work in their local environment.  We don't want individual developer versions of these files to get checked in, but for many of them it is good to have a starting point with common coding conventions and the like.  We capture these pre-configured files in a new `configTemplates` directory (**N.B.** `templates` is not an option because that is used by IntelliJ for other purposes) and provide some new tasks that will install or reset these files.  These tasks will also, automatically use the right path separator in the runConfigurations files.

We need to test this in the update scenario and the start-from-scratch scenarios:

**Start from Scratch**
- `git clone https://github.com/LabKey/server cleanRoot`
- `cd cleanRoot/server`
- `git clone https://github.com/LabKey/testAutomation`
- `cd modules # now in cleanRoot/server/modules`
- `git clone https://github.com/LabKey/platform`
- `cd ../../ # now in cleanRoot`
- `gradlew deployApp`
- `gradlew ijConfigure`
- `gradlew ijWorkspaceSetup`
- Open cleanRoot project in IntelliJ

**Updating**

- Shut down IntelliJ
- Make a copy of your `.idea` directory (just in case)
- Stash the current changes in the `.idea` directory  using `git stash push -- .idea/` from the root of your LabKey enlistment
- `git pull`
- `./gradlew ijConfigure`
- Open your existing project in IntelliJ

**Local Configuration**
After opening your new or existing project in IntelliJ, and after IntelliJ has had a chance to index and sync with the new settings, you should verify and possibly adjust some settings to fit your local installation:
- Check the Gradle window to see if the projects show up.  If not, do a Gradle refresh in the Gradle window.
- Go to Preferences -> Appearance & Behavior -> Path Variables and set a value for NODE_PATH that will point to the node binary used by the build (e.g.,`/Users/susanhert/Development/labkey/root/build/.node/node/bin/node`)
- Possibly Edit the LabKey Dev config to fix the module classpath (This should not be necessary if you are running the version 2021.1 or later of IntelliJ.)
- Possibly Edit the LabKey Production config to fix the module classpath (This should not be necessary if you are running the version 2021.1 or later of IntelliJ.)
- Possibly Edit the LabKeyEmbedded Dev config to fix the module classpath (This should not be necessary if you are running the version 2021.1 or later of IntelliJ.)
- Go to Preferences -> Languages & Frameworks -> Typescript and verify the Node interpreter settings.  The interpreter should be pointing to the node version that you set as the NODE_PATH variable.  If not, or if you need to change this at some point, you can do that here.  

**To Verify**
In each case, the following should be verified

- [x] Gradle projects show up (Does this require a gradle refresh?)
- [x] `.tsx` files should not show red squiggles 
- [x] Labkey Dev configuration works
- [x] LabKey Production configuration works
- [x] LabKeyEmbedded_Dev configuration works (You'll need to uncomment the `useEmbeddedTomcat` and `useLocalBuild` properties within your root-level `build.gradle` file and do a `gradlew deployApp` for it to find the requisite project.)
- [x] Running tests from within IntelliJ works
- [x] Java files do not have errors related to newer versions of Java syntax
- [x] IntelliJ does not complain about a missing SDK.

#### Changes
* Remove checked in version of .idea files that are changed by IntelliJ
* Add starting versions for these IntelliJ files that incorporate our styling preferences and basic configurations
* Add gradle tasks to do the setup from the template files.


.idea/configTemplates/compiler.xml
```diff
@@ -0,0 +1,29 @@
...
+      <entry name="?*.html" />
+      <entry name="?*.gif" />
+      <entry name="?*.png" />
+      <entry name="?*.jpeg" />
+      <entry name="?*.jpg" />
+    </wildcardResourcePatterns>
+    <bytecodeTargetLevel target="14" />
+  </component>
+  <component name="TypeScriptCompiler">
+    <option name="nodeInterpreterTextField" value="$PROJECT_DIR$/.node/node/bin/node" />
```
labkey-susanh: For Windows users, this will need to be updated within IntelliJ.

.idea/runConfigurations/templates/LabKeyEmbedded_Dev.xml
```diff
@@ -1,14 +1,15 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="LabKeyEmbbedded_Dev" type="SpringBootApplicationConfigurationType" factoryName="Spring Boot">
-    <module name="server.server.embedded.main" />
+    <module name="org.labkey.root.server.embedded.main" />
```
labkey-susanh: Is there a better option here so this becomes more generally applicable?

.idea/runConfigurations/templates/LabKey_Production.xml
```diff
@@ -1,17 +1,12 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="LabKey Production" type="Application" factoryName="Application">
-    <extension name="coverage" enabled="false" merge="false" sample_coverage="true" runner="idea" />
+    <option name="ALTERNATIVE_JRE_PATH" value="labkey" />
+    <option name="ALTERNATIVE_JRE_PATH_ENABLED" value="true" />
     <option name="MAIN_CLASS_NAME" value="org.apache.catalina.startup.Bootstrap" />
-    <option name="VM_PARAMETERS" value="-Dcatalina.base=&quot;./&quot; -Dcatalina.home=&quot;./&quot; -Djava.io.tmpdir=&quot;./temp&quot; -Ddevmode=false -Dsun.io.useCanonCaches=false -Xmx2G -classpath &quot;./bin/*;$APPLICATION_HOME_DIR$/lib/idea_rt.jar&quot;" />
+    <module name="root.server.modules.platform.api.main" />
```
labkey-susanh: Is there a better option here so this becomes more generally applicable?

.idea/runConfigurations/templates/LabKey_Dev.xml
```diff
@@ -0,0 +1,10 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="LabKey Dev" type="Application" factoryName="Application">
+    <option name="MAIN_CLASS_NAME" value="org.apache.catalina.startup.Bootstrap" />
+    <module name="root.server.modules.platform.api" />
```
labkey-susanh: Is there a better option here so this becomes more generally applicable?

.idea/configTemplates/misc.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectRootManager" version="2" project-jdk-name="labkey" project-jdk-type="JavaSDK"/>
```
labkey-tchad, line 3: Do we want to define a default language level? `languageLevel="JDK_14"` is what I have locally.

.idea/configTemplates/vcs.xml
```diff
@@ -20,5 +30,9 @@
   </component>
   <component name="VcsDirectoryMappings">
     <mapping directory="" vcs="Git" />
+    <mapping directory="$PROJECT_DIR$/server/modules/commonAssays" vcs="Git" />
+    <mapping directory="$PROJECT_DIR$/server/modules/platform" vcs="Git" />
+    <mapping directory="$PROJECT_DIR$/server/modules/tutorialModules" vcs="Git" />
+    <mapping directory="$PROJECT_DIR$/server/testAutomation" vcs="Git" />
```
labkey-tchad: Do we want to include 'tutorialModules' by default? The other three are what we specify in the [dev machine setup](https://www.labkey.org/Documentation/wiki-page.view?name=devMachine#files) docs so I think they make sense to have.

.idea/configTemplates/sqldialects.xml
```diff
@@ -0,0 +1,26 @@
...
+    <file url="file://$PROJECT_DIR$/server/modules/platform/pipeline/resources/schemas/dbscripts/sqlserver" dialect="TSQL" />
+    <file url="file://$PROJECT_DIR$/server/modules/platform/query/resources/schemas/dbscripts/postgresql" dialect="PostgreSQL" />
+    <file url="file://$PROJECT_DIR$/server/modules/platform/query/resources/schemas/dbscripts/sqlserver" dialect="TSQL" />
+    <file url="file://$PROJECT_DIR$/server/modules/platform/search/resources/schemas/dbscripts/postgresql" dialect="PostgreSQL" />
+    <file url="file://$PROJECT_DIR$/server/modules/platform/search/resources/schemas/dbscripts/sqlserver" dialect="TSQL" />
+    <file url="file://$PROJECT_DIR$/server/modules/platform/study/resources/schemas/dbscripts/postgresql" dialect="PostgreSQL" />
+    <file url="file://$PROJECT_DIR$/server/modules/platform/study/resources/schemas/dbscripts/sqlserver" dialect="TSQL" />
+    <file url="file://$PROJECT_DIR$/server/modules/platform/survey/resources/schemas/dbscripts/postgresql" dialect="PostgreSQL" />
+    <file url="file://$PROJECT_DIR$/server/modules/platform/survey/resources/schemas/dbscripts/sqlserver" dialect="TSQL" />
```
labkey-tchad, line 24: Could we have the `ijConfigure` task (or a separate task) set the SQL dialects for all the modules currently checked out?

.idea/runConfigurations/templates/LabKeyEmbedded_Dev.xml
labkey-tchad: I think this is based off of the root project's name in gradle. I have `rootProject.name="labkey-server"` defined in my `settings.gradle` file in several enlistments and they all have `<module name="labkey-server.server.embedded.main" />` for their embedded run configuration.
This also seems to have the same effect for the non-embedded run configurations.
Note that IntelliJ will generate a file (`.idea/.name`) with that name in it, which we would want to ignore or commit.

.idea/runConfigurations/templates/LabKey_Dev.xml
```diff
@@ -0,0 +1,10 @@
...
+    <option name="PROGRAM_PARAMETERS" value="start" />
+    <option name="VM_PARAMETERS" value="-Dcatalina.base=&quot;./&quot; -Dcatalina.home=&quot;./&quot; -Djava.io.tmpdir=&quot;./temp&quot; -Ddevmode=true -ea -Dsun.io.useCanonCaches=false -Xmx2G -classpath &quot;./bin/*PATH_SEPARATOR$APPLICATION_HOME_DIR$/lib/idea_rt.jar&quot;" />
+    <option name="WORKING_DIRECTORY" value="/usr/local/apache-tomcat-9.0.33" />
```
labkey-tchad: Can/should this use CATALINA_HOME?

.idea/runConfigurations/templates/LabKey_Production.xml
```diff
@@ -1,17 +1,12 @@
...
-    <option name="WORKING_DIRECTORY" value="file://$CATALINA_HOME$" />
-    <option name="ALTERNATIVE_JRE_PATH_ENABLED" value="false" />
-    <option name="ALTERNATIVE_JRE_PATH" />
-    <option name="ENABLE_SWING_INSPECTOR" value="false" />
-    <option name="ENV_VARIABLES" />
-    <option name="PASS_PARENT_ENVS" value="true" />
-    <module name="api" />
-    <envs />
+    <option name="VM_PARAMETERS" value="-Dcatalina.base=&quot;./&quot; -Dcatalina.home=&quot;./&quot; -Djava.io.tmpdir=&quot;./temp&quot; -Ddevmode=false -Dsun.io.useCanonCaches=false -Xmx2G -classpath &quot;./bin/*PATH_SEPARATOR$APPLICATION_HOME_DIR$/lib/idea_rt.jar&quot;" />
+    <option name="WORKING_DIRECTORY" value="/usr/local/apache-tomcat-9.0.33" />
```
labkey-tchad: CATALINA_HOME?

.idea/configTemplates/sqldialects.xml
labkey-tchad: Didn't realize the tasks aren't defined in the gradle plugin. This seems significantly more complicated than the other new tasks.

.idea/configTemplates/misc.xml
labkey-susanh: I had that originally, but IntelliJ seems to do the right thing based on the JavaSDK, so I left it out.

.idea/configTemplates/vcs.xml
labkey-susanh: Probably not. I'll take that out.

.idea/configTemplates/sqldialects.xml
labkey-susanh: Yeah, to do it for all modules checked out (as opposed to in the settings file) would be considerably slower and since you can set it when you visit a file and then just go about your business without need to do that again, think I'll leave it.

.idea/runConfigurations/templates/LabKey_Dev.xml
labkey-susanh: Ah, good thought.  I'll try that.

labkey-susanh: Unfortunately, this does not work. IntelliJ does not know anything about CATALINA_HOME (or other environment variables, apparently).  Strangely, if this value is defined to use `$CATALINA_HOME$` or, following the example from the file that is deleted `file://$CATALINA_HOME$`, it finds the oldest tomcat I have installed, so somewhere it knows how to look for something tomcat-like, but it's definitely not using my CATALINA_HOME environment variable.   Any ideas?

labkey-adam: Might be obvious, but IntelliJ has "Path Variables" (Settings -> Appearance & Behavior) that you can define within IntelliJ. I don't think they interact with environment variables. In the past, the run configurations have been defined to use these path variables. I believe IntelliJ will map these both directions (i.e., if you save LabKey_Dev it will reverse match the CATALINA_HOME path variable and replace the path with the variable name on save). I have CATALINA_HOME defined as a path variable, but it's mapped to an old value. I stopped bothering to update these values a while back.

labkey-adam: https://www.jetbrains.com/help/idea/2021.1/absolute-path-variables.html

labkey-susanh: Thanks, @labkey-adam.  That's where I have this old tomcat version set.  We do still recommend this route for setting up your dev machine (https://www.labkey.org/Documentation/wiki-page.view?name=devMachine), so I'll update these files to use `$CATALINA_HOME$`.

labkey-adam: Agreed... definitely better than hard-coded. Probably works well for those who don't upgrade Tomcat once a month...

labkey-adam (approved):

labkey-tchad (approved):

PR #67, opened 2021-04-08
Update Gradle plugins for Gradle 7 support and a few other dependencies

labkey-susanh:
#### Rationale
Gradle 7 is on its way and we want to use it when it comes out so we can start building with Java 16.  We've had to make a few minor changes to our gradle plugins to work with the latest Gradle 7 release candidate.  We also take this opportunity to update some dependency versions.

#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/122
* https://github.com/LabKey/platform/pull/2169

#### Changes
* Turn on build caching by default
* Update some dependency versions

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,7 +3,7 @@
 # Uncomment the following line to turn on Gradle's file watching, which may improve efficiency
 # https://blog.gradle.org/introducing-file-system-watching
 #org.gradle.vfs.watch=true
-# Uncomment the following line to turn on Gradle's build cache, which will improve efficiency
+# This controls Gradle's build cache, which improves efficiency.
 # https://docs.gradle.org/current/userguide/build_cache.html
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -5,7 +5,7 @@
 #org.gradle.vfs.watch=true
 # Uncomment the following line to turn on Gradle's build cache, which will improve efficiency
```
labkey-jeckels:
```suggestion
# This controls Gradle's build cache, which improves efficiency
```

labkey-jeckels (approved):

labkey-adam (changes requested):
Not a change for this repo, but don't we need a corresponding jars.txt update in platform?

labkey-adam (approved):

PR #75, opened 2021-05-07
Update to Gradle 7

labkey-susanh:
#### Rationale
Gradle 7 is the latest and greatest


labkey-adam (approved):

labkey-adam:
Looks like itn-tslk-modules/overview needs an adjustment (`compile` -> `implementation`?). Let me know if you'd like me to make that change.

PR #76, opened 2021-05-10
Publish Libraries to deploy into Tomcat

labkey-tchad:
#### Rationale
LabKey distributions include several jars that get deployed to `$CATALINA_HOME/lib`. The gradle plugin currently pulls these jars from the `tomcatJars` configuration of the server project. This new task (`:server:tomcatLibsZip`) allows a standalone module build to pull compatible libraries from Artifactory to produce a distribution.

#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/123

#### Changes
* Add `:server:tomcatLibsZip` task and publish its output to Artifactory

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -33,6 +33,14 @@ dependencies
    xsdDoc "docflex:docflex-xml-re:${docflexXmlReVersion}"
 }
 
+project.tasks.register("tomcatLibsZip", Zip) {
+    Zip zip ->
+        zip.group = "Build"
+        zip.description = "produce jar file with jars to be deployed in \$CATALINA_HOME/lib"
+        zip.from project.configurations.tomcatJars.getAsFileTree()
+        zip.archiveBaseName.set("tomcat-libs")
+        zip.destinationDirectory = project.file(project.labkey.explodedModuleLibDir)
+}
 
 configurations
         {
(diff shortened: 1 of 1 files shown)


server/build.gradle
```diff
@@ -154,10 +162,25 @@ project.publishing {
             }
         }
 
+        tomcatJars(MavenPublication) {
+            groupId = 'org.labkey.api'
```
labkey-tchad: Not sure what the appropriate group is. I also considered `org.labkey` and `org.labkey.build`.

labkey-susanh: I think `org.labkey.build` makes sense.

server/build.gradle
```diff
@@ -33,6 +33,14 @@ dependencies
    xsdDoc "docflex:docflex-xml-re:${docflexXmlReVersion}"
 }
 
+project.tasks.register("tomcatLibsZip", Zip) {
+    Zip zip ->
+        zip.group = "Build"
```
labkey-susanh: Could use `GroupNames.BUILD` here instead.

labkey-susanh (approved):
Looks good.  I would change the group name of the artifact, though.  `org.labkey.api` doesn't make as much sense to me as the other options.

PR #91, opened 2021-06-16
Merge changes from 21.6.0

labkey-tchad:
#### Rationale
Merging changes forward from 21.6.0


labkey-susanh (approved):
I just merged my changes to update various versions, which included an update gradlePlugins 1.28.0, so you'll likely need to merge that in here before merging.

PR #112, opened 2021-07-26
Update plugins version and add new purgeNpmAlphaVersions task

labkey-susanh:
#### Rationale
The alpha versions of our NPM packages never get removed unless we do that manually, and currently the manual method for doing this within the Artifactory web application requires finding and clicking on each version separately.  Even filtering to the proper versions within the Artifactory UI is not really possible.  Here we add a new Gradle task to do that purging.

#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/130

#### Changes
* Add task `purgeNpmAlphaVersions`


cnathe (approved):
Looks good. I used the command to remove the alpha packages for the fb_auditCompQueryModel, and it worked great!

PR #142, opened 2021-10-27
bump LabKey version

labkey-klum:
#### Rationale
increment the LabKey version number for the 21.7.8 maintenance release


labkey-tchad:
This is automated now!
https://github.com/LabKey/server/pull/139

PR #153, opened 2021-11-26
Compile with Java 16 source & target

labkey-adam:
#### Rationale
Allow Java 16 syntax

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2813

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -245,7 +245,7 @@ slf4jLog4j12Version=1.7.32
 slf4jLog4jApiVersion=1.7.32
 
 
-springBootVersion=2.6.1.RELEASE
+springBootVersion=2.6.0
 # This MUST match the Tomcat version dictated by springBootVersion
 springBootTomcatVersion=9.0.55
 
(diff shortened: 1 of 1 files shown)


labkey-susanh (approved):

gradle.properties
```diff
@@ -245,9 +245,9 @@ slf4jLog4j12Version=1.7.32
 slf4jLog4jApiVersion=1.7.32
 
 
-springBootVersion=2.2.13.RELEASE
+springBootVersion=2.6.1.RELEASE
```
labkey-adam: @labkey-jeckels Did this work locally? I think they've stopped using the `.RELEASE` suffix on their versions.

labkey-adam: Also looks like 2.6.1 isn't available for every artifact we use. 2.6.0 might be good for now.

labkey-jeckels: It did, but only because I had failed to uncomment the "useEmbeddedTomcat" property. Trying again...

PR #156, opened 2021-12-08
Gradle 7.3.1

labkey-susanh:
#### Rationale
It's the latest version. I don't know that we'll be affected by the fixes in this version, but seems best to keep current.


labkey-tchad (approved):
DRT failures are from a TeamCity bug

Sigmonia (approved):

labkey-adam (approved):

labkey-adam:
@labkey-tchad So this TeamCity bug has nothing to do with compiling to Java 16 byte codes? If that's the case, then the TeamCity upgrade shouldn't block the Java 16 PRs, right?

labkey-tchad:
@labkey-adam Yeah, its probably safe to merge the Java 16 PRs. I'm not sure why the DRTs are hitting this error but other suites are not.

PR #158, opened 2021-12-10
Upgrade to NPM LTS current release for compatibility with M1/arm64 CPUs

labkey-jeckels:
#### Rationale
New MacBooks use a new chipset. NPM 14.x didn't publish builds that were compiled for the new architecture, but newer versions do. 16.x is their current LTS release

**PR Against Develop:** https://github.com/LabKey/server/pull/161

#### Changes
* Switch from NPM 14.7.1 to 16.3.1

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -65,7 +65,7 @@ versioningPluginVersion=1.1.0
 # will be downloaded and used. If not set, the existing local installations will be use
 # The version of npm corresponds to the given version of node
 npmVersion=6.14.13
-nodeVersion=14.17.1
+nodeVersion=16.13.1
 nodeRepo=https://nodejs.org/dist
 # Directory in a project's build directory where the node binary will be placed
 nodeWorkDirectory=.node


labkey-susanh (commented):
Seems innocuous enough, but please be sure to run tests that exercise node on both Linux and Windows (e.g., the Biologics or SM Selenium and Jest tests).

gradle.properties
```diff
@@ -65,7 +65,7 @@ versioningPluginVersion=1.1.0
 # will be downloaded and used. If not set, the existing local installations will be use
 # The version of npm corresponds to the given version of node
 npmVersion=6.14.13
-nodeVersion=14.17.1
+nodeVersion=16.13.1
```
labkey-nicka, line 68: Why are we doing this against 21.7? Is a client developer using an M1 Macbook? Can they just change it locally? 

This seems much better to go into `develop` and let it bake rather than destabilizing a release or downstream releases (i.e. v21.11).

gradle.properties
```diff
@@ -65,7 +65,7 @@ versioningPluginVersion=1.1.0
 # will be downloaded and used. If not set, the existing local installations will be use
 # The version of npm corresponds to the given version of node
 npmVersion=6.14.13
```
labkey-nicka, line 67: We tend to try and keep the `node` and `npm` versions linked as these are often paired together across major releases. See https://nodejs.org/en/download/releases/ for a mapping.

This isn't to say we couldn't unlink them but it's not a configuration that is normally installed when installing Node.

labkey-nicka (changes requested):
I'm not sure the binary distribution updates warrant a release update.

PR #169, opened 2021-12-15
Use embedded config for declaring dependency on Springboot jar file

labkey-susanh:
#### Rationale
After the update to springBoot version 2.6, in conjunction with adding support for Java 16, the default jar file is now being created from the embedded project, and that default jar is what is being added to distributions.  We introduce a separate configuration that has the `bootJar` attached as an artifact and use that for declaring dependencies.

#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/139

#### Changes
* Try to prevent the creation of the "plain" jar file from the embedded project
* Add an `embedded` configuration to attach the `bootJar` file to
* Publish the output of `bootJar` not `jar`.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -57,7 +57,7 @@ windowsProteomicsBinariesVersion=1.0
 # The current version numbers for the gradle plugins.
 artifactoryPluginVersion=4.21.0
 gradleNodePluginVersion=3.0.1
-gradlePluginsVersion=1.30.3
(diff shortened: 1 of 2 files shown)


labkey-tchad (approved):
Looking good so far: https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Internal_Embedded_SampleManagerEmbeddedPostgres/1639318
The DRT failures are from a TeamCity bug (we will apply the fix later this week).

Sigmonia (approved):

PR #188, opened 2021-12-30
Use new property name

labkey-adam:
#### Rationale
New property name


labkey-susanh (approved):

labkey-susanh:
We will also need to update this value in the labkey.xml in the `distributionResources` directory in `gradlePlugins`

labkey-adam:
Will do

PR #202, opened 2022-01-31
Upgrade a couple dependencies and use preferred property name

labkey-adam:
#### Rationale
We're behind on a couple dependencies and we like using the most current property name. We're also sticklers about accuracy in our comments.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -208,7 +208,7 @@ kaptchaVersion=2.3
 
 log4j2Version=2.17.1
 
-mysqlDriverVersion=8.0.27
+mysqlDriverVersion=8.0.28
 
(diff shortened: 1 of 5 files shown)


labkey-susanh (approved):

labkey-susanh:
Don't forget the update for `jars.txt`

labkey-adam:
I don't think there's anything to update... MySQL is handled automatically and SpringBoot isn't annotated anywhere that I can see

labkey-susanh:
> I don't think there's anything to update... MySQL is handled automatically and SpringBoot isn't annotated anywhere that I can see

Yes, you're right. Perhaps we should have those in our credits somewhere, but we don't currently.

PR #210, opened 2022-02-28
Issue 44781: Update application.properties and LabKeyServer to use "encryptionKey"

ankurjuneja:
#### Rationale
This PR replaces the usage of deprecated property master encryption key with encryption key in embedded tomcat.

#### Related Pull Requests
* https://github.com/LabKey/Dockerfile/pull/9


labkey-jeckels (changes requested):
1. This PR is pointing at develop when the issue is against 22.3. Since we just branched, this should get a new FB name with a 22.3_fb_ prefix and a new PR.
2. Do we care about being backwards compatible with any existing application.properties that may exist in the real world? Perhaps we're OK *only* supporting the new name going forward, but then we need to let people know that they'll need to refresh.

ankurjuneja:
1. Created https://github.com/LabKey/server/pull/214
2. I think we can let people know that they'll need to do pickDB task to refresh the application.properties.

PR #212, opened 2022-03-01
ContentSecurityPolicyFilter

labkey-matthewb:
#### Rationale
Customers who want to use Content-Security-Policy headers don't have an easy way to do this.  I could not find a filter included with Tomcat to set this header.  This creates a class that an admin can choose to use if they want to set this header.

#### Related Pull Requests
* 

#### Changes
*


labkey-jeckels (approved):
Adding this seems good to me. We'll separately determine if we want use it to apply a CSP on various instances.

PR #213, opened 2022-03-01
Issue 44947: Make log rotation file naming consistent across log types

labkey-jeckels:
#### Rationale
It's confusing when we rotate logs in opposite directions

#### Changes
* Always use .1 as the most recently rotated file, and .2, .3, etc as the older copies

### DIFF
--- a/webapps/log4j2.xml
+++ b/webapps/log4j2.xml
@@ -53,7 +53,7 @@
                 <OnStartupTriggeringPolicy />
                 <SizeBasedTriggeringPolicy size="100 MB"/>
             </Policies>
-            <DefaultRolloverStrategy max="3"/>
+            <DefaultRolloverStrategy max="3" fileIndex="min" />
         </RollingFile>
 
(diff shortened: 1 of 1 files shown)


labkey-susanh (commented):
Is this change going to cause any issues with existing log files?

labkey-jeckels:
@labkey-susanh if a server already has .1, .2, etc log files, until they're deleted by the rotation we'll have a mix of numbering systems within a single log file type. For example, sorting from newest to oldest might give:

labkey.log
labkey.log.1
labkey.log.2
labkey.log.7
labkey.log.6
labkey.log.5
labkey.log.4
labkey.log.3

but they'll eventually all rotate through to the new preferred ordering. I'll add a release note that nobody will really understand.

labkey-susanh:
As long as the notes aren't understandable, the goal of consistency seems good to me.

labkey-susanh (approved):

PR #223, opened 2022-03-22
Issue 45087: Exclude log4j dependency in all configurations

labkey-susanh:
#### Rationale
We don't want to include the old log4j library in any of our builds, so we explicitly exclude it from all configurations.  

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3159

#### Changes
* Exclude log4j from all configurations.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -161,6 +161,8 @@ allprojects {
                     force "org.slf4j:slf4j-api:${slf4jLog4jApiVersion}"
                     // force some newer versions than are brought in by mondrian (et al.)
                     force "xerces:xercesImpl:${xercesImplVersion}"
+                    // force invalid version 0 so the build will fail if we inadvertently include the old log4j with its possible vulnerabilities again
+                    force "log4j:log4j:0"
                     force "org.apache.logging.log4j:log4j-core:${log4j2Version}"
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):
Thanks for trying this out - glad it worked

build.gradle
```diff
@@ -69,6 +69,9 @@ allprojects {
                 driver
                 utilities
             }
+    configurations.all {
+        exclude group: "log4j", module:"log4j"
+    }
```
Sigmonia, line 75: Just to make sure I'm understanding: unlike the prior iterations/commits this will block loading of log4j (instead of loading a dummy version) and fail at the module with some kind of Not Found error?

Rather than loading the insecure version.

Sigmonia (approved):

build.gradle
labkey-susanh: Correct.  If there is indeed a dependency on log4j, there will likely be a "ClassNotFound" error of some sort. Though the cause of the build error may not be as immediately obvious, this seems preferable to me because (a) it's much less of a hack (b) it achieves what we actually want, so we won't fail a build if there's a transitive dependency that doesn't actually get used (c) it's a once-and-for-all kind of fix.

PR #224, opened 2022-03-25
Close streams in ExplodedModule.copyFile

labkey-tchad:
#### Rationale
`ExplodedModule.copyFile` doesn't close all of its streams when complete. I suspect this is behind some failures we see in the moduleEditor tests on Windows (they pass with this change anyway). I've applied some changes from `FileUtil.copyFile`, which this method claims to be a copy of. It now closes the input and output streams as well as acquiring a `FileLock`, which I'm not familiar with but assume has a purpose.

#### Changes
* Include streams in try-with-resources declarations

### DIFF
--- a/server/bootstrap/src/org/labkey/bootstrap/ExplodedModule.java
+++ b/server/bootstrap/src/org/labkey/bootstrap/ExplodedModule.java
@@ -15,11 +15,24 @@
  */
 package org.labkey.bootstrap;
 
-import java.io.*;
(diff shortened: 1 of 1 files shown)


labkey-tchad:
Unfortunately, this doesn't actually fix the moduleEditor failure (it had just been muted). Might still be worth updating though.

labkey-matthewb (approved):

PR #234, opened 2022-04-18
Declare mavenCentral repository explicitly as first repository

labkey-susanh:
#### Rationale
We plan to update our Artifactory configuration so it no longer proxies for certain remote repositories. This requires that builds that require these remote repositories declare these repositories themselves.  We put mavenCentral as the first repository so it won't query our Artifactory instance when looking for external dependencies.

#### Related Pull Requests
* 

#### Changes
* Stop publishing `buildInfo`
* Explicitly declare mavenCentral as first repository to check

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ buildscript {
     repositories {
         mavenCentral()
         maven {
-            url "${artifactory_contextUrl}/plugins-release-no-mc"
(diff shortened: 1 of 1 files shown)


labkey-tchad (approved):

labkey-susanh:
Changes will be pulled froward from 21.3

PR #236, opened 2022-04-19
Put mavenCentral last again so we pick up our custom version of the radeox lib

labkey-susanh:
#### Rationale
Though we have updated to a new version of radeox in develop (or, actually on 2/16/2022) and don't rely on our own version of this, for earlier releases we do use one that is published in the `ext-release-local` repository in artifactory, so we need to find that one first.  We can do the switching of ordering in develop, but can merge this change forward in the meantime.

#### Related Pull Requests
#235

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -98,7 +98,6 @@ allprojects {
 
     repositories
             {
-                mavenCentral()
                 // this if statement is necessary because the TeamCity artifactory plugin overrides
                 // the repositories but does not use these artifactory_ urls.  For others who are
                 // developing or building, you do need to specify the three artifactory_ properties
@@ -138,6 +137,7 @@ allprojects {
                         }
                     }
                 }
+                mavenCentral()
             }
     configurations.all
             {


labkey-jeckels:
@labkey-susanh I may be misinterpreting your comment here, but we do still rely on our own Radeox build that I uploaded to Artifactory in February:

https://artifactory.labkey.com/ui/packages/gav:%2F%2Fradeox:radeox/1.0.4?name=radeox&type=packages&activeTab=builds

https://github.com/LabKey/platform/pull/3049

https://github.com/LabKey/platform/pull/3056

labkey-susanh:
> @labkey-susanh I may be misinterpreting your comment here, but we do still rely on our own Radeox build that I uploaded to Artifactory in February:
> 
> https://artifactory.labkey.com/ui/packages/gav:%2F%2Fradeox:radeox/1.0.4?name=radeox&type=packages&activeTab=builds
> 
> [LabKey/platform#3049](https://github.com/LabKey/platform/pull/3049)
> 
> [LabKey/platform#3056](https://github.com/LabKey/platform/pull/3056)

@labkey-jeckels Ah, yes, you're right there, but the difference is that there isn't a 1.0.4 version currently out in the wild, or at least not available from mavenCentral.  So, we'll get our version regardless of the ordering. We could keep this ordering, even in develop to avoid inadvertently picking up a version we don't want, since I'm not entirely sure how much the order switch will help us one way or the other in the end, and not sure at this point how to assess that, but if we believe there's a significant difference in traffic we could, I think, make this switch in develop.

labkey-jeckels:
Yes, if I'm following along correctly, I agree this change should be a safe one. I just want to be sure that my excellent and exciting efforts on Radeox bug fixes don't get accidentally dropped.

Hopefully 21.3 and other older releases that might pick up this change won't be generating much traffic anyway.

labkey-tchad (approved):

PR #264, opened 2022-06-13
Issue 37040: Test Microsoft JDBC driver for SQL Server

labkey-jeckels:
#### Rationale
The official Microsoft JDBC driver has some nice capabilities beyond jTDS, including higher availability.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3440

#### Changes
* Update properties
* Make MS driver the default choice via pickMsSql


labkey-susanh (commented):
Is there value in adding a pickSqlServer Gradle task as well?  It could be easily added via a new gradlePlugin release or added directly into the `server/build.gradle` file.

labkey-jeckels:
> Is there value in adding a pickSqlServer Gradle task as well? It could be easily added via a new gradlePlugin release or added directly into the `server/build.gradle` file.

@labkey-susanh that sounds nice while we're supporting both. Maybe we should call it pickJTDS/jtds.properties instead?

labkey-tchad (approved):
I like the idea of changing it to `jtds.properties`. Either way works for TeamCity though.

labkey-susanh:
> I like the idea of changing it to `jtds.properties`. Either way works for TeamCity though.

👍

labkey-susanh:
> > Is there value in adding a pickSqlServer Gradle task as well? It could be easily added via a new gradlePlugin release or added directly into the `server/build.gradle` file.
> 
> @labkey-susanh that sounds nice while we're supporting both. Maybe we should call it pickJTDS/jtds.properties instead?

@labkey-jeckels Do you want to do the gradlePlugin update or do you want to me do it?

labkey-tchad:
> @labkey-jeckels Do you want to do the gradlePlugin update or do you want me to do it?

I'll do it. Seems like an easy change to make wile waiting for my build to finish.

labkey-tchad:
https://github.com/LabKey/gradlePlugin/pull/149

PR #279, opened 2022-07-21
Update artifactory URL for 22.7

Sigmonia:
#### Rationale
Artifactory migration

#### Related Pull Requests
* TBD

#### Changes
* Change artifactory_contextUrl gradle property

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,7 +12,7 @@ buildscript {
             url "https://plugins.gradle.org/m2"
         }
         maven {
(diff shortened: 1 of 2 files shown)


labkey-susanh (approved):

labkey-susanh (commented):
I would kind of like to update references to the repositories with the "-no-proxy" suffix to their counterparts without that suffix (e.g., plugins-release-no-proxy becomes plugins-release).  That would allow me to get rid of the `-no-proxy` virtual repos in Artifactory, but if that seems too risky at this point, don't worry about it.

PR #285, opened 2022-07-26
Gradle 7.5 and remove no-proxy suffix from artifact repositories

labkey-susanh:
#### Rationale
Latest is greatest and it's good to tidy things up when we can.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -127,9 +127,9 @@ allprojects {
                                 username = artifactory_user
                                 password = artifactory_password
                             }
-                        }
-                        authentication {
-                            basic(BasicAuthentication)
+                            authentication {
+                                basic(BasicAuthentication)
+                            }
                         }
                     }
                     maven {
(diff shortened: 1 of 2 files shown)


labkey-tchad (approved):

build.gradle
```diff
@@ -119,31 +119,31 @@ allprojects {
...
                             credentials {
                                 username = artifactory_user
                                 password = artifactory_password
                             }
-                            authentication {
-                                basic(BasicAuthentication) // enable preemptive authentication to get around https://www.jfrog.com/jira/browse/RTFACT-4434
-                            }
+                        }
+                        authentication {
+                            basic(BasicAuthentication)
```
labkey-tchad: Community builds/tests are failing on TeamCity because they run without credentials.

labkey-tchad (changes requested):

build.gradle
labkey-susanh: Oops. I put that back in the wrong place... Fixed now.

labkey-tchad (approved):

Sigmonia (approved):

PR #289, opened 2022-07-27
Issue 45969: Update spring-expression to 5.3.22

labkey-jeckels:
#### Rationale
We want to adopt some of the fixes in spring-expression before we do a full Spring upgrade later, in 22.11

#### Changes
* Force version to 5.3.22 while keeping the rest of Spring at the older release

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -252,7 +252,7 @@ allprojects {
                     force "io.netty:netty-codec-http:${nettyVersion}"
 
                     force "org.springframework:spring-core:${springVersion}"
-                    force "org.springframework:spring-expression:5.3.17"
+                    force "org.springframework:spring-expression:5.3.22"
 
                     dependencySubstitution {
                         // Because the client api artifact name is not the same as the directory structure, we use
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
@labkey-susanh I initially tried to force the versions in the API module but couldn't figure out a functional syntax. Happy to move into API if you prefer and can help with guidance.

build.gradle
```diff
@@ -251,6 +251,9 @@ allprojects {
                     force "io.netty:netty-common:${nettyVersion}"
                     force "io.netty:netty-codec-http:${nettyVersion}"
 
+                    force "org.springframework:spring-core:${springVersion}"
+                    force "org.springframework:spring-expression:5.3.22"
```
labkey-susanh, line 258: Perhaps add a comment about why this forcing is done (and why the versions are different).  Are we forcing both because forcing spring-expression then tries to update spring-core?

labkey-susanh (approved):
@labkey-jeckels It's best to do this forcing here rather than in API because then it applies to all modules that might bring in this dependency.

PR #304, opened 2022-08-18
Issue 45345: Optimize default DB connection pool config

labkey-jeckels:
#### Rationale
Reviewing performance data, we're spending a surprising amount of time validating DB connections before using them.

#### Changes
* Switch from validating DB connections when borrowing from the pool to idle times
* Be more prescriptive for the desired size of the pool - initial and idle

### DIFF
--- a/webapps/labkey.xml
+++ b/webapps/labkey.xml
@@ -8,7 +8,6 @@
         driverClassName="@@jdbcDriverClassName@@"
         url="@@jdbcURL@@"
         maxTotal="20"
-        initialSize="3"
         maxIdle="3"
         minIdle="1"
         maxWaitMillis="120000"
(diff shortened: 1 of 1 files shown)


labkey-stuartm:
#https://github.com/LabKey/syseng-chef-server/pull/400 
Mirror changes in chef lk_appserver cookbook

labkey-adam (approved):

labkey-adam:
Are the [FATAL: database "bt3" does not exist] bootstrap errors related to your changes?

labkey-stuartm (approved):

labkey-jeckels:
@labkey-adam yes, turns out that specifying an initialSize causes the connection pool to throw a SQLException when initializing due to the lack of a DB, breaking bootstrapping. I've backed out that part of the change, which was less important than the other edits.

PR #310, opened 2022-08-27
Upgrade several dependencies

labkey-adam:
#### Rationale
Latest is greatest

#### Changes
* Upgrade javax.servlet-api 3.0 -> 4.0.1
* Upgrade PostgreSQL JDBC driver 42.4.1 -> 42.5.0
* Upgrade Spring Boot 2.7.2 -> 2.7.3
* Upgrade quartz 2.1.7 -> 2.3.2


server/build.gradle
```diff
@@ -23,7 +23,7 @@ dependencies
         {
             tomcatJars  "com.sun.mail:jakarta.mail:${javaMailVersion}"
             tomcatJars  "com.sun.activation:jakarta.activation:${activationVersion}" // JavaMail dependency (starting with Java 10), so it must be copied to <tomcat>/lib
-            remotePipelineJars "javax.servlet:servlet-api:${servletApiVersion}"
+            remotePipelineJars "javax.servlet:javax.servlet-api:${servletApiVersion}"
```
labkey-susanh, line 26: We reference the servlet api in the [Jsp gradle plugin](https://github.com/LabKey/gradlePlugin/blob/f85049cc9928eeeeac65699aa9a956ce17a915f2/src/main/groovy/org/labkey/gradle/plugin/Jsp.groovy#L91) as well, so I'm guessing we'll need to update that plugin as well. (Unfortunately, I didn't leave much of a comment here about the forced version, so it might be worth trying removing the forced version to see if we are doing the right thing already.)

labkey-adam: @labkey-susanh I can create an FB that updates the plugin... though I'll need some help coordinating the publishing, etc. I am puzzled by one thing: AFAICT, `javax.servlet:servlet-api:3.0` doesn't exist in Maven Central. So did we push this to our Artifactory?

labkey-susanh: @labkey-adam  TeamCity will do the publishing of the snapshot versions for you, and when you're ready to merge I can do the publishing of the release version.

As for the servlet-api version, yes we published a version locally. It's [here](https://labkey.jfrog.io/artifactory/ext-release-local/javax/servlet/servlet-api/3.0/).

labkey-jeckels (approved):

PR #333, opened 2022-10-25
Bump Spring Boot & Tomcat versions. Remove unused json-simple property.

labkey-adam:
#### Rationale
Latest Spring Boot & Tomcat

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -95,6 +95,9 @@ apacheDirectoryVersion=1.0.3
 #Transitive dependency of Apache directory: 2.0.18 contains some regressions
 apacheMinaVersion=2.0.19
 
+# Keep in sync with springBootTomcatVersion below
+apacheTomcatVersion=9.0.68
+
 # (mothership) -> json-path -> json-smart -> accessor-smart
 # (core) -> graalvm
 # tika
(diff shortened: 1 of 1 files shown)


gradle.properties
```diff
@@ -264,9 +260,11 @@ slf4jLog4j12Version=1.7.33
 # this version is forced for compatibility with api, LDK, and workflow
 slf4jLog4jApiVersion=1.7.33
 
-springBootVersion=2.7.3
+springBootVersion=2.7.5
 # This MUST match the Tomcat version dictated by springBootVersion
-springBootTomcatVersion=9.0.65
+springBootTomcatVersion=9.0.68
+# Usually kept in sync with spring boot Tomcat version above
+apacheTomcatVersion=9.0.68
```
labkey-jeckels: I understand wanting to group these values within the file, but obviously that breaks the alphabetical ordering. @labkey-susanh do you have preferences?

labkey-jeckels (approved):

gradle.properties
labkey-susanh: My preference is still alphabetical (except for those few bits at the top). It seems a recipe for no organization at all if we start to try to group related things together.  In the end, most everyone will use an editor to search for the version of interest, so as long as the comment is there it's not a big deal (comment might reference the actual property name, though).

labkey-adam: Yeah, I figured I'd be called out for this. I knew about both properties, but still managed to let them get out-of-sync at some point. But I'll switch it back and add a comment.

PR #377, opened 2022-12-19
Upgrade SQL Server and Oracle JDBC drivers

labkey-adam:
#### Rationale
Newest versions of SQL Server (11.2.3) and Oracle (21.8.0.0) JDBC drivers

Note: 11.2.3 fixes a regression in SQLException getSQLState() and getErrorCode() return values that also appeared in 12.1.0-preview, see https://github.com/microsoft/mssql-jdbc/issues/2015. Microsoft is planning to release 12.2.0 at the end of January. Merging this PR now to get us the latest stable version; we'll revisit 12.x when Microsoft makes it available.


labkey-susanh (commented):
I don't know that approving a draft version carries any weight, and definitely don't want this merged before the actual release version for the mssql driver comes through, so ping me again when that happens.

labkey-susanh (approved):

PR #397, opened 2023-02-06
Preparing for Gradle 8 - update "with" to "using"

labkey-susanh:
#### Rationale
It seems I may have missed a memo somewhere about a syntax change for dependency substitution. As of Gradle 8, instead of `substitute module ... with ...`, it's now `substitute module ... using ...`

#### Related Pull Requests
* https://github.com/LabKey/cloud/pull/122

#### Changes
* "with" becomes "using"


build.gradle
```diff
@@ -262,25 +262,25 @@ allprojects {
                             substitute module('org.labkey:labkey-client-api') with project(BuildUtils.getRemoteApiProjectPath(gradle))
```
labkey-tchad: Does the `labkey-client-api` substitution need to be updated?
```suggestion
                            substitute module('org.labkey:labkey-client-api') using project(BuildUtils.getRemoteApiProjectPath(gradle))
```

labkey-tchad (approved):
Missed one.

build.gradle
labkey-susanh: Yes, thanks. Wasn't building that locally.

PR #415, opened 2023-03-15
Teach IntelliJ not to import the wrong Logger and StringUtils classes

labkey-adam:
#### Rationale
There are quite a few classes named `StringUtils` and `Logger`, most of which we don't want to auto-import.


labkey-jeckels (approved):
Thanks for tracking down the right file to commit

labkey-adam:
> Thanks for tracking down the right file to commit

@labkey-jeckels FYI: Trick was to switch the "scope" for these excluded classes from IDE to Project. That caused IntelliJ to create an XML file in the expected place. "IDE" settings must be saved with the IntelliJ installation or something.

PR #422, opened 2023-03-22
Add mavenLocal repository to buildscripts

labkey-tchad:
#### Rationale
Including `mavenLocal` allows one to more easily test gradle plugin changes.

#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/171

#### Changes
* Add mavenLocal() repository to buildscripts
* Update gradle plugin to improve task configuration performance

### DIFF
--- a/gradle/settings/all.gradle
+++ b/gradle/settings/all.gradle
@@ -5,25 +5,13 @@ buildscript {
         }
         if (gradlePluginsVersion.contains("SNAPSHOT"))
         {
+            mavenLocal()
             maven {
                 url "${artifactory_contextUrl}/plugins-snapshot-local"
             }
 
         }
     }
-    // For testing changes to the gradle plugins, uncomment the dependencies block below and reference the location
-    // of your gradlePlugin enlistment.  Then comment out the following dependencies
-    // block.  You will need to build the gradlePlugins jar file manually to pick up new changes.
-//    dependencies {
-//        classpath files("../gradlePlugin/build/libs/gradlePlugin-${gradlePluginsVersion}.jar")
-//        classpath "org.apache.commons:commons-lang3:${commonsLang3Version}"
-//        classpath "org.apache.commons:commons-text:${commonsTextVersion}"
-//        classpath "commons-io:commons-io:${commonsIoVersion}"
-//        classpath "org.apache.httpcomponents:httpclient:${httpclientVersion}"
-//        classpath "org.json:json:${orgJsonVersion}"
-//        classpath "com.fasterxml.jackson.core:jackson-databind:${jacksonDatabindVersion}"
-//        classpath "org.ajoberstar.grgit:grgit-gradle:${grgitGradleVersion}"
-//    }
     dependencies {
         classpath "org.labkey.build:gradlePlugins:${gradlePluginsVersion}"
     }
--- a/gradle/settings/parameters.gradle
+++ b/gradle/settings/parameters.gradle
@@ -10,4 +10,5 @@ gradle.ext.remoteApiProjectPath=":remoteapi:labkey-api-java"
 gradle.ext.jdbcApiProjectPath=":remoteapi:labkey-api-jdbc"
 gradle.ext.schemasProjectPath=":server:modules:platform:api"
 gradle.ext.testProjectPath=":server:testAutomation"
+gradle.ext.minificationProjectPath=":server:minification"
 gradle.ext.nodeBinProjectPath=":" // The npm and node executables downloaded into this project's build directory will be the target of the links created by the synlinkNode task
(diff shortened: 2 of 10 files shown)


gradle/settings/parameters.gradle
```diff
@@ -10,4 +10,5 @@ gradle.ext.remoteApiProjectPath=":remoteapi:labkey-api-java"
 gradle.ext.jdbcApiProjectPath=":remoteapi:labkey-api-jdbc"
 gradle.ext.schemasProjectPath=":server:modules:platform:api"
 gradle.ext.testProjectPath=":server:testAutomation"
+gradle.ext.minificationProjectPath=":server:minification"
```
labkey-susanh, line 13: This isn't really necessary since this is the default value anyway and I don't expect anyone is going to change this.

gradle/settings/all.gradle
```diff
@@ -5,25 +5,13 @@ buildscript {
...
-    // block.  You will need to build the gradlePlugins jar file manually to pick up new changes.
-//    dependencies {
-//        classpath files("../gradlePlugin/build/libs/gradlePlugin-${gradlePluginsVersion}.jar")
-//        classpath "org.apache.commons:commons-lang3:${commonsLang3Version}"
-//        classpath "org.apache.commons:commons-text:${commonsTextVersion}"
-//        classpath "commons-io:commons-io:${commonsIoVersion}"
-//        classpath "org.apache.httpcomponents:httpclient:${httpclientVersion}"
-//        classpath "org.json:json:${orgJsonVersion}"
-//        classpath "com.fasterxml.jackson.core:jackson-databind:${jacksonDatabindVersion}"
-//        classpath "org.ajoberstar.grgit:grgit-gradle:${grgitGradleVersion}"
```
labkey-susanh, line 25: This should mean we can remove some of the comments and properties in the `gradle.properties` file that were added to support this block. Look for comments like this: `# used in gradle plugins; included here for use in testing changes locally`.

settings.gradle
```diff
@@ -1,15 +1,17 @@
...
+//        maven {
+//            url "${artifactory_contextUrl}/plugins-release"
+//        }
+//        if (gradlePluginsVersion.contains("SNAPSHOT") || versioningPluginVersion.contains("SNAPSHOT")) {
+//            mavenLocal()
+//            maven {
+//                url "${artifactory_contextUrl}/plugins-snapshot-local"
+//            }
+//        }
+//    }
```
labkey-susanh: I'm surprised if it's not necessary. My read of things is that Gradle will not look in our custom repository for plugins unless we configure them. Perhaps the declaration of the repositories in the `buildScript` block applies here as well, but that seems counterintuitive to me.

gradle/settings/parameters.gradle
labkey-tchad: I was just adding it to be consistent with the other `*ProjectPath` parameters.

settings.gradle
labkey-tchad: It seems to fall back to the `buildscript` repositories if this is removed. Mangling the URLs in `pluginManagement.repositories` does produce an error, so it isn't ignored entirely.

I'm not sure the `pluginManagement.plugins` block is doing anything at all though. I can remove plugins, reference non-existent plugins, reference non-existent versions, or change plugins to `apply true`; all with no apparent negative impact on my build.
I think adding `gradlePlugins` to the classpath below overrides any version you might specify in `pluginManagement`.

labkey-susanh: The `pluginManagement` block is definitely the more modern way to do things, but I'm unclear on the benefits, other than knowing it is the modern syntax. If we take out the classpath dependency and leave the `pluginManagement` block (including a use of `mavenLocal`), does it still work well, even with a locally published version of the plugins?

labkey-tchad: If I remove the classpath dependency, it fails because it can't find `BuildUtils`.
It seems like custom types like BuildUtils or our various task types can only be brought in via classpath. The docs imply that you can pull in custom types via plugin but the only [example of importing a custom task type](https://docs.gradle.org/current/userguide/custom_tasks.html#sec:using_your_task_class_in_another_project) does so using classpath.
It looks like we might be able to [wrap `BuildUtils` in an extension](https://docs.gradle.org/current/userguide/custom_gradle_types.html#nested_objects) and expose that through a plugin but I don't think it's worth the trouble.
Having our plugin jar on the buildscript classpath seems to override everything referencing those plugins in the `pluginManagement` block.
I fiddled with it some more and found a scenario where the `pluginManagement` block matters. If we want to pull in the 'versioning' plugin without adding it to the classpath, pluginManagement has to have the correct repositories.

labkey-susanh (approved):
Thanks for doing this clean up, Trey.

PR #423, opened 2023-03-22
Issue 47533: Update third-party dependencies

labkey-jeckels:
#### Rationale
We want to be on the latest versions of these important dependencies

#### Changes
* Bump Commons BeanUtils
* Add reference to external repo for Mondrian artifacts


labkey-susanh (approved):
Looks like there are a few other repos that need versions updated to make the verifyDependencies task work well.

PR #478, opened 2023-05-12
Enable access log for embedded Tomcat

labkey-tchad:
#### Rationale
The tomcat access log is occasionally quite useful. I've had it [enabled](https://docs.spring.io/spring-boot/docs/current/reference/html/application-properties.html#application-properties.server.server.tomcat.accesslog.enabled) locally for embedded tomcat for quite a while.
Explicitly setting `server.tomcat.basedir` makes it so our logs don't end up in a temp directory.

#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/187

#### Changes
* Enable access log for embedded Tomcat
* Set `server.tomcat.basedir` for embedded Tomcat


server/configs/application.properties
```diff
@@ -51,3 +51,9 @@ management.endpoints.web.exposure.include=*
 
 # Don't show the Spring banner on startup
 spring.main.banner-mode=off
+
+# Enable tomcat access log
+server.tomcat.basedir=@@pathToServer@@/build/deploy/embedded
```
labkey-jeckels: I'm not sure where this properties file is used. If it's used outside of dev machines, will this be a desirable path?

labkey-jeckels (approved):

server/configs/application.properties
labkey-tchad: It will not be desireable, in fact. I made a gradle plugin update to treat this like the other properties with `@@pathToServer@@`; to leave it commented out unless you're on a dev machine.

labkey-jeckels: Great

ankurjuneja (approved):

server/configs/application.properties
```diff
@@ -72,14 +72,24 @@ spring.main.banner-mode=off
 #jsonaccesslog.condition-if=attributeName
 #jsonaccesslog.condition-unless=attributeName
 
-# Uncomment below properties (csp.policy and csp.disposition) to enable Content Security Policy (CSP) headers
+# Uncomment one of the below properties (csp.report or csp.enforce) to enable Content Security Policy (CSP) headers
```
ankurjuneja: one or both can be uncommented.

PR #479, opened 2023-05-12
Don't download proteomics binaries by default

labkey-tchad:
#### Rationale
Most tests don't need these. Stop wasting bandwidth downloading them for every single build on TeamCity.

#### Related Pull Requests
* N/A

#### Changes
* Don't download proteomics binaries unless requested

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -49,7 +49,7 @@ configurations
         }
 configurations.binaries.setDescription("Utility and proteomics binaries")
 
-if (!project.hasProperty('excludeProteomicsBinaries'))
+if (project.hasProperty('includeProteomicsBinaries'))
 {
     dependencies
     {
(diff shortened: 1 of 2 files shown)


server/build.gradle
```diff
@@ -49,7 +49,7 @@ configurations
         }
 configurations.binaries.setDescription("Utility and proteomics binaries")
 
-if (!project.hasProperty('excludeProteomicsBinaries'))
+if (project.hasProperty('includeProteomicsBinaries'))
```
labkey-susanh, line 52: You could do this the opposite way by adding the exclude property to the `gradle.properties` file and leaving this as is, which would mean no need to update docs or potentially communicate the switch to anyone.  I guess this way forces someone to make the decision again to choose them and might make it easier to remove altogether, but please do update the [documentation](https://www.labkey.org/Documentation/wiki-page.view?name=gradleTips) and at least let the support team know of the change.

labkey-susanh (approved):

server/build.gradle
labkey-tchad: I was going to do that but wanted to retain the ability to control the property without modifying any files. It was either a new property or make `excludeProteomicsBinaries` take a value, which would also require documentation. This seemed the least invasive change, presuming that a majority of developers don't need the binaries.

PR #490, opened 2023-05-26
Synchronize Google API versions

labkey-jeckels:
#### Rationale
There's an incompatibility between the versions of googleApiClientVersion and the Google Drive and Google Calendar libraries.

#### Changes
* Upgrade library versions and move one dependency used in single module to that repo

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -145,6 +145,7 @@ fopVersion=2.7
 
 googleApiClientVersion=2.2.0
 # Force latest for consistency
+googleErrorProneAnnotationsVersion=2.18.0
 googleHttpClientGsonVersion=1.42.2
 googleHttpClientVersion=1.43.2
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -143,11 +143,11 @@ flyingsaucerVersion=R8
 # Apache FOP -- linked to Apache Batik version above
 fopVersion=2.7
 
-googleApiServicesCalendarVersion=v3-rev411-1.25.0
-googleApiClientVersion=2.0.0
+googleApiClientVersion=2.2.0
 # Force latest for consistency
+googleErrorProneAnnotationsVersion=2.18.0
```
labkey-susanh, line 148: The name doesn't instill confidence, but is appropriate.

labkey-susanh (approved):

labkey-martyp (approved):

PR #497, opened 2023-06-06
Configure repository content for faster searching for dependencies

labkey-susanh:
#### Rationale
By default, every repository that is declared is searched for dependencies, which can be wasteful when we know that some repositories will never contain certain dependencies.  Gradle has introduced ways to [configure repositories](https://docs.gradle.org/current/userguide/declaring_repositories.html#sec:repository-content-filtering) to filter the content that is served from them, reducing search time and traffic to the repo server when looking for a dependency.  Might as well take advantage of that. 

#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/177
* https://github.com/LabKey/tutorialModules/pull/154
* https://github.com/LabKey/labkey-api-r/pull/94
* https://github.com/LabKey/labkey-api-jdbc/pull/36
* https://github.com/LabKey/labkey-api-java/pull/58

#### Changes
* Designate repos as either snapshot-only or release-only
* Filter to known groups for tool repository and snapshot repository
* Use parallel garbage collection

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -103,7 +103,9 @@ allprojects {
                     maven {
                         url "${artifactory_contextUrl}/ext-tools-local"
                         content {
-                            includeGroup "org.labkey.tools"
+                            includeGroup "org.labkey.tools.windows"
+                            includeGroup "org.labkey.tools.linux"
+                            includeGroup "org.labkey.tools.osx"
                         }
                     }
                     maven {
--- a/build.gradle
+++ b/build.gradle
@@ -102,9 +102,11 @@ allprojects {
                 {
                     maven {
                         url "${artifactory_contextUrl}/ext-tools-local"
+                        content {
+                            includeGroup "org.labkey.tools"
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -131,7 +138,17 @@ allprojects {
...
                             }
                         }
+                        mavenContent {
+                            snapshotsOnly()
+                        }
+                        content {
+                            includeGroup "org.labkey"
+                            includeGroup "org.labkey.api"
+                            includeGroup "org.labkey.module"
+                            includeGroup "org.maccosslab"
```
labkey-tchad: The libraries in `org.maccosslab` are from an abandoned project (no updates since 2019 and not referenced by anything). Not sure its worth including here.
https://github.com/LabKey/goomba-workflow-framework
https://github.com/LabKey/maccosslab-workflow

labkey-tchad (approved):

PR #529, opened 2023-07-22
Issue 48281: ability to set additional database connection pool properties

Sigmonia:
#### Rationale
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=48281

#### Changes
* Added new managed props to applications.properties and the ContextProperties classes

PR #534, opened 2023-07-26
Make a better check if application property is not set

Sigmonia:
#### Rationale
Defaults are not applied correctly after https://github.com/LabKey/server/pull/529

#### Related Pull Requests
* https://github.com/LabKey/server/pull/529

#### Changes
* add checks for String.isBlank

### DIFF
--- a/server/embedded/src/org/labkey/embedded/LabKeyServer.java
+++ b/server/embedded/src/org/labkey/embedded/LabKeyServer.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.jar.JarFile;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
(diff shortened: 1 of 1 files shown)


server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -161,23 +161,23 @@ private List<ContextResource> getDataSourceResources(ContextProperties props) th
                     dataSourceResource.setProperty("username", props.getUsername().get(i));
 
                     String maxTotal = props.getMaxTotal().get(i);
```
labkey-jeckels: Based on the error Alan shared, `props.getMaxTotal()` is returning null:

```
org.springframework.context.ApplicationContextException:
Unable to start web server; nested exception is java.lang.NullPointerException: 
Cannot invoke "java.util.List.get(int)" because the return value of "org.labkey.embedded.LabKeyServer$ContextProperties.getMaxTotal()" is null
```

I think we need null checks on the list itself, and checks on its length too.

labkey-jeckels (changes requested):

server/embedded/src/org/labkey/embedded/LabKeyServer.java
Sigmonia: I added a null check for the list & caught the IndexExceptions and return the default value for both. Could possibly use a Map instead of a List which would allow key checking, but not sure of any secondary ramifications of the type change.

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -160,32 +160,38 @@ private List<ContextResource> getDataSourceResources(ContextProperties props) th
...
+            {
+                if (propList == null)
+                {
+                    logger.debug(String.format("%1$s property was not provided, using default", propName));
+                    return defaultValue;
+                }
+
+                try
+                {
+                    String val = propList.get(index);
```
labkey-jeckels: It's better to check the list size instead of catching the exception. Just like we should check for nulls instead of catching NullPointerException. Otherwise looks good.

labkey-jeckels (approved):

server/embedded/src/org/labkey/embedded/LabKeyServer.java
Sigmonia: Changed to use Map.

Using a List means that any props provided in secondary resources need to also be provided for primary (and/or previously configured resources). The Map seems to bind correctly, and correctly keys the property values.

labkey-jeckels (approved):

PR #541, opened 2023-08-08
Bump HTSJDK and Picard versions

labkey-jeckels:
#### Rationale
There's a new HTSJDK version. Among other updates, it no longer depends on MJSON.

#### Changes
* Update to the latest HTSJDK
* Drop the suppression on the now-defunct MJSON dependency

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -217,7 +217,7 @@ allprojects {
                     force "xalan:xalan:${xalanVersion}"
                     // genotyping brings in a much older version of this, so we force a newer version for compatibility
                     force "org.apache.commons:commons-compress:${commonsCompressVersion}"
-                    // Transitive dependency of commons-compress -- Transitive dependency com.github.samtools:htsjdk:2.14.3 uses much older version
+                    // Transitive dependency of commons-compress -- Transitive dependency com.github.samtools:htsjdk which references an older version
                     force "org.tukaani:xz:${tukaaniXZVersion}"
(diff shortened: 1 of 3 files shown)


bbimber:
hi @labkey-jeckels i will comment on the other PR, but what do you think about adding a picardVersion variable here? It's not used in core, but if we expect htsjdk+picard to be kept in step, it would make future PRs simpler? I would still keep the comment about syncing them in this file

bbimber:
@labkey-jeckels : this is not related, but given you are editing gradle.properties could i request you add luceneVersion to this file and drop it from the search module? This will let jbrowse share it.

UPDATE: please disregard this. has already been done

labkey-jeckels:
> 

I'm not opposed to pulling the picardVersion here, but am inclined to keep it as-is. Generally we're trying to keep the centralized list as lean as possible. It also lets you be self-serve if you just want to bump Picard instead of needing to loop us in. Although admittedely as this PR shows, there are enough shared dependencies that it's hard to do any of these changes totally in isolation.

labkey-adam (approved):

bbimber:
> > 
> 
> I'm not opposed to pulling the picardVersion here, but am inclined to keep it as-is. Generally we're trying to keep the centralized list as lean as possible. It also lets you be self-serve if you just want to bump Picard instead of needing to loop us in. Although admittedely as this PR shows, there are enough shared dependencies that it's hard to do any of these changes totally in isolation.

OK, that's fine

Sigmonia (approved):

PR #548, opened 2023-08-15
Force ant version

Sigmonia:
#### Rationale
https://teamcity.labkey.org/viewLog.html?buildId=2584515&buildTypeId=LabKey_Trunk_Premium_InternalSuites_OwaspDependencyCheck

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4675

#### Changes
* force version of Ant build dependency

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -88,13 +88,14 @@ npmWorkDirectory=.node
 # These are the libraries that are used by our "base" modules
 # (api, bootstrap, schemas, remoteapi). Other modules that
 # have extensive other external dependencies should have a local
-# gradle.properties file to declare these version numbers.
+# gradle.properties file to declare these version numbers. Naming
+# convention is <library name>Version camel-cased, i.e. "jacksonVersion".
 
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -94,6 +94,8 @@ activationVersion=1.2.2
 
 annotationsVersion=15.0
 
+apacheAntVersion=1.10.13
```
labkey-susanh: I see we've become a little inconsistent with naming convention, but usually this would just be the name the artifact (`ant`) followed by `version` (so `antVersion`).

labkey-susanh (approved):

gradle.properties
labkey-jeckels: Can we capture our desired convention in a comment in the file itself? It won't guarantee compliance, of course, but it's more likely than leaving it unstated.

Sigmonia: Updated the variable, and added a comment at the top of the section to include the naming convention.

PR #551, opened 2023-08-18
Use library substitution to prevent Eigenbase conflicts

Sigmonia:
#### Rationale
Use substitution option to prevent library conflicts

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4681
* https://github.com/LabKey/platform/pull/4684

#### Changes
* change from conflict exclusion to substitution

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -317,6 +317,11 @@ allprojects {
 
                         // This library was renamed
                         substitute module('org.codehaus.woodstox:woodstox-core-asl') using module("com.fasterxml.woodstox:woodstox-core:${woodstoxCoreVersion}")
+
+                        // Avoid conflicts with declared dependencies on the newer versions from 'net.hydromatic'
+                        substitute module('eigenbase:eigenbase-properties:1.1.2') using module("net.hydromatic:eigenbase-properties:${eigenbasePropertiesVersion}")
+                        substitute module('eigenbase:eigenbase-resgen:1.3.1') using module("net.hydromatic:eigenbase-resgen:${eigenbaseResgenVersion}")
+                        substitute module('eigenbase:eigenbase-xom:1.3.5') using module("net.hydromatic:eigenbase-xom:${eigenbaseXomVersion}")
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -317,6 +317,11 @@ allprojects {
 
                         // This library was renamed
                         substitute module('org.codehaus.woodstox:woodstox-core-asl') using module("com.fasterxml.woodstox:woodstox-core:${woodstoxCoreVersion}")
+
+                        // Avoid conflicts with declared dependencies on the newer versions from 'net.hydromatic'
+                        substitute module('eigenbase:eigenbase-properties:1.1.2') using module("net.hydromatic:eigenbase-properties:${eigenbasePropertiesVersion}")
```
labkey-jeckels: Do we need the version number on the old coordinates or can we drop them?

labkey-jeckels (commented):
I'll defer to Susan on the full approval and recommended way to reference the old variant

build.gradle
labkey-susanh: The version number is not required from Gradle, so the version is needed only if we think a substitution of some older or new version with `net.hyromatic` library would cause problems. I'm guessing the older coordinates have been abandoned, but can't find any text that actually says that. I think removing the version should be safe. We can restore it if needed.

gradle.properties
```diff
@@ -147,6 +147,9 @@ datadogVersion=0.108.1
 dom4jVersion=2.1.4
 
 ehcacheCoreVersion=2.6.8
+eigenbasePropertiesVersion=1.1.6
+eigenbaseResgenVersion=1.3.7
+eigenbaseXomVersion=1.3.6
```
labkey-susanh: Looks like there's also a 1.3.7 version of eigenbase-xom, with fewer CVEs.

labkey-susanh (changes requested):

labkey-susanh (approved):

PR #568, opened 2023-09-11
Add option to log document indexing on a per-category basis to a sepa…

labkey-adam:
…rate log file

#### Rationale
Additional logging, filtered to specific document categories, will be useful in tracking down aggressive indexing scenarios

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4732

### DIFF
--- a/webapps/log4j2.xml
+++ b/webapps/log4j2.xml
@@ -136,6 +136,21 @@
             <DefaultRolloverStrategy fileIndex="min" />
         </RollingFile>
 
+        <RollingFile name="INDEXED_DOCUMENTS"
+                     fileName="${sys:labkey.log.home}/labkey-indexed-documents.log"
+                     append="true"
+                     bufferedIO="false"
+                     filePattern="${sys:labkey.log.home}/labkey-indexed-documents.log.%i">
+            <!-- d=datetime m=message n=newline -->
+            <PatternLayout>
+                <Pattern>%d{ISO8601} %m%n</Pattern>
+            </PatternLayout>
+            <Policies>
+                <SizeBasedTriggeringPolicy size="1 MB"/>
+            </Policies>
+            <DefaultRolloverStrategy max="3" fileIndex="min" />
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):
I don't know where we want to capture this data, but note that this won't work for our new container-based deployments, which use a different Log4J config and stream logging via stdout instead of log files.

@labkey-stuartm @labkey-willm

labkey-willm:
> I don't know where we want to capture this data, but note that this won't work for our new container-based deployments, which use a different Log4J config and stream logging via stdout instead of log files.
> 
> @labkey-stuartm @labkey-willm

Doesn't the config just fine tune that particular logger? If they're being surfaced under 'org.labkey', I've got all that going to the container's stdout already. Or, if I need to add that full api path specifically, let me know and I can do so.

labkey-adam:
> > I don't know where we want to capture this data, but note that this won't work for our new container-based deployments, which use a different Log4J config and stream logging via stdout instead of log files.
> > @labkey-stuartm @labkey-willm
> 
> Doesn't the config just fine tune that particular logger? If they're being surfaced under 'org.labkey', I've got all that going to the container's stdout already. Or, if I need to add that full api path specifically, let me know and I can do so.

Not sure how to answer this question in terms of a "logger", but this adds a new RollingFile Appender that writes to a separate file to ease analysis. Plan is to use this locally and on labkey.org to analyze cases where we're indexing more aggressively than we expect. It's possible we'll enable this type of logging on a client's instance (in cloud or on-premise), though I suspect that'll be rare.

labkey-jeckels:
My reading is that because this sets `additivity="false"` debug-level messages wouldn't also get routed to the separate console logger. It would only go to the separate file. It won't blow up, but we won't be able to get these messages in container-based deployments.

Maybe that's fine, but we've had ongoing disconnects in the Log4J config files and expected behavior so I thought it was worth mentioning.

PR #570, opened 2023-09-19
Update `org.apache.commons:commons-compress` to 1.24.0

labkey-tchad:
#### Rationale
Fixing error: https://teamcity.labkey.org/test/529818193974818065?currentProjectId=LabKey_237Release_Internal_InternalSuites&expandTestHistoryChartSection=true

#### Related Pull Requests
* N/A

#### Changes
* Update `org.apache.commons:commons-compress` to 1.24.0

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -124,8 +124,7 @@ commonsBeanutilsVersion=1.9.4
 commonsCollectionsVersion=3.2.2
 commonsCollections4Version=4.4
 commonsCodecVersion=1.15
-# sync with version Tika ships
-commonsCompressVersion=1.23.0
+commonsCompressVersion=1.24.0
 commonsDbcpVersion=1.4
 commonsDiscoveryVersion=0.2
 commonsDigesterVersion=1.8.1
--- a/gradle.properties
+++ b/gradle.properties
@@ -124,6 +124,7 @@ commonsBeanutilsVersion=1.9.4
 commonsCollectionsVersion=3.2.2
 commonsCollections4Version=4.4
 commonsCodecVersion=1.15
+# sync with version Tika ships
 commonsCompressVersion=1.24.0
 commonsDbcpVersion=1.4
 commonsDiscoveryVersion=0.2


labkey-adam (approved):

labkey-adam:
Tika does directly rely on Commons Compress and our current Tika version (and the most recent release) specifies 1.23.0. I suspect there will be no problem with 1.24.0, but the Tika link is still relevant.

For example: https://github.com/apache/tika/blob/2.8.0/tika-parent/pom.xml

labkey-tchad:
> Tika does directly rely on Commons Compress and our current Tika version (and the most recent release) specifies 1.23.0. I suspect there will be no problem with 1.24.0, but the Tika link is still relevant.
> 
> For example: https://github.com/apache/tika/blob/2.8.0/tika-parent/pom.xml

Oh, thank you. I didn't see it in the gradle dependency tree because we have the tika libraries set to `transitive = false`

labkey-adam:
> > Tika does directly rely on Commons Compress and our current Tika version (and the most recent release) specifies 1.23.0. I suspect there will be no problem with 1.24.0, but the Tika link is still relevant.
> > For example: https://github.com/apache/tika/blob/2.8.0/tika-parent/pom.xml
> 
> Oh, thank you. I didn't see it in the gradle dependency tree because we have the tika libraries set to `transitive = false`

I believe that's because, left to its own devices, Tika brings in the entire world. We used to include it all (running Tika in its own class loader to prevent conflicts) but now we curate a list of parsers (therefore file formats) that we care about.

Sigmonia:
Looks like there is a new Tika version (2.9.0), but that would likely mean bumping a few other libraries that should also be sync'ed. Also, I'm not sure if it actually brings the new desired version or not, Maven Central's new search isn't cooperating with its dependency graph.

labkey-adam:
> Looks like there is a new Tika version (2.9.0), but that would likely mean bumping a few other libraries that should also be sync'ed. Also, I'm not sure if it actually brings the new desired version or not, Maven Central's new search isn't cooperating with its dependency graph.

We should upgrade Tika, but no need to link it with this change

labkey-tchad:
> Looks like there is a new Tika version (2.9.0), but that would likely mean bumping a few other libraries that should also be sync'ed. Also, I'm not sure if it actually brings the new desired version or not, Maven Central's new search isn't cooperating with its dependency graph.

Tika 2.9.0 doesn't update commons-compress: https://mvnrepository.com/artifact/org.apache.tika/tika-parsers/2.9.0

Sigmonia:
Thanks Trey that's what I was looking for. I should have also figured it out, given their respective release dates too.

Sigmonia (approved):

PR #587, opened 2023-10-11
force netty to 4.1.100.Final for CVE-2023-4586

labkey-willm:
#### Rationale
force netty to 4.1.100.Final for CVE-2023-4586

#### Related Pull Requests
* 

#### Changes
*


labkey-jeckels (approved):
This is a good change and, as noted, clears the failure.

I imagine it may resurface as Netty doesn't plan to change their default behavior until version 5.x.

PR #588, opened 2023-10-11
Issue 48868: Change Encryption Key doesn't work with embedded Tomcat

labkey-jeckels:
#### Rationale
We need to pipe through each parameter from the embedded Tomcat configuration into its equivalent spot in the server context

#### Changes
* Propagate `context.oldEncryptionKey` and a couple of other, unrelated parameters we haven't yet added
* Add to the `application.properties` template

### DIFF
--- a/server/configs/application.properties
+++ b/server/configs/application.properties
@@ -29,9 +29,9 @@ context.validationQuery[0]=SELECT 1
 
 #context.webAppLocation=@@pathToServer@@/build/deploy/labkeyWebapp
 context.encryptionKey=@@encryptionKey@@
-context.oldEncryptionKey=oldKey
-context.requiredModules=rM
-context.pipelineConfig=pC
+#context.oldEncryptionKey=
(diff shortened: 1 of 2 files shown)


server/configs/application.properties
```diff
@@ -29,6 +29,9 @@ context.validationQuery[0]=SELECT 1
 
 #context.webAppLocation=@@pathToServer@@/build/deploy/labkeyWebapp
 context.encryptionKey=@@encryptionKey@@
+context.oldEncryptionKey=oldKey
+context.requiredModules=rM
+context.pipelineConfig=pC
```
labkey-susanh: Should these be commented out by default?

labkey-susanh (approved):

server/configs/application.properties
labkey-jeckels: Yes, they should. Thanks. Now they are.

PR #591, opened 2023-10-16
Adopt Tomcat 9.0.82

labkey-jeckels:
#### Rationale
There are some bug fixes in the latest Tomcat release that we want to pick up ahead of its adoption in the upcoming Spring Boot 2.7.17. We can back out the forcing to the version when the new Spring Boot is available

#### Changes
* Force to the latest Tomcat

### DIFF
--- a/server/embedded/build.gradle
+++ b/server/embedded/build.gradle
@@ -31,7 +31,38 @@ dependencies {
     implementation "org.springframework.boot:spring-boot-starter-web:${springBootVersion}"
     implementation "org.springframework.boot:spring-boot-starter-actuator:${springBootVersion}"
     implementation "org.springframework.boot:spring-boot-starter-validation:${springBootVersion}"
-    
+
+    // Force to the latest Tomcat version until Spring Boot 2.7.17 is released and we can adopt it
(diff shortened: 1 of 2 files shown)


labkey-adam (approved):

server/embedded/build.gradle
```diff
@@ -31,7 +31,38 @@ dependencies {
...
+    }
+    implementation('org.apache.tomcat:tomcat-jsp-api') {
+        version {
+            strictly "${springBootTomcatVersion}"
+        }
+    }
+//
+//    force ":${springBootTomcatVersion}"
+//    force "org.apache.tomcat.embed:tomcat-embed-el:"
+//    force "org.apache.tomcat.embed:tomcat-embed-websocket:${springBootTomcatVersion}"
```
labkey-susanh: Can probably get rid of these since I think the above syntax is preferable here.

labkey-susanh (approved):

server/embedded/build.gradle
labkey-jeckels: I cleaned them up

PR #595, opened 2023-10-17
Upgrade many dependencies

labkey-adam:
#### Rationale
Keeping current


labkey-susanh (approved):
All seems good to me, but the tests will surely tell.

PR #597, opened 2023-10-18
Support Content-Security-Policy-Report-Only

labkey-matthewb:
#### Rationale
Support Content-Security-Policy-Report-Only
Allows for easier testing of a new CSP for instance

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/server/bootstrap/src/org/labkey/filters/ContentSecurityPolicyFilter.java
+++ b/server/bootstrap/src/org/labkey/filters/ContentSecurityPolicyFilter.java
@@ -32,7 +32,11 @@ import java.util.Enumeration;
             frame-ancestors 'self' ;
           </param-value>
         </init-param>
-      </filter>
+        <init-param>
+          <param-name>disposition</param-name>
+          <param-value>enforce</param-value>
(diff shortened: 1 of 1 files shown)


server/bootstrap/src/org/labkey/filters/ContentSecurityPolicyFilter.java
```diff
@@ -68,9 +75,16 @@ public void init(FilterConfig filterConfig) throws ServletException
                 policy = s;
                 nonceSubstIndex = policy.indexOf(NONCE_SUBST);
             }
+            else if ("disposition".equals(paramName))
```
labkey-jeckels: If we're case insensitive for the values do we want to be for the parameter name too?

labkey-jeckels (approved):

server/bootstrap/src/org/labkey/filters/ContentSecurityPolicyFilter.java
labkey-matthewb: Hmm, why not.

PR #598, opened 2023-10-19
Adopt the latest Spring Boot and Tomcat releases

labkey-jeckels:
#### Rationale
Latest is greatest (except when Tomcat 9.0.81 was the latest)

#### Changes
* Keep Spring Boot and Tomcat versions in sync


labkey-adam (changes requested):
I suggest some changes

PR #607, opened 2023-10-25
updating netty version to 4.1.100 in suppression for CVE-2023-4586

labkey-willm:
#### Rationale
updating netty version to 4.1.100 in suppression for CVE-2023-4586

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/dependencyCheckSuppression.xml
+++ b/dependencyCheckSuppression.xml
@@ -276,7 +276,7 @@
         <notes><![CDATA[
    file name: netty-handler-4.1.100.Final.jar
    ]]></notes>
-        <packageUrl regex="true">^pkg:maven/io\.netty/netty\-handler@.*$</packageUrl>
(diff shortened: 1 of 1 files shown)


labkey-jeckels (changes requested):
The OWASP check is still failing for all of the other Netty JARs that we pull in. I think need to adjust the packageUrl to catch more of them. I haven't tested, but something like `^pkg:maven/io\.netty/netty.*$` might do it

labkey-willm:
> The OWASP check is still failing for all of the other Netty JARs that we pull in. I think need to adjust the packageUrl to catch more of them. I haven't tested, but something like `^pkg:maven/io\.netty/netty.*$` might do it

so changed. Sorry about the premature PR request. I saw all green on the tests but I guess the OWASP hadn't yet been added/run.

labkey-jeckels (approved):
Looks good!

PR #621, opened 2023-11-14
Replace `buildDir` with `layout.buildDirectory` and add an `excludedModules` property

labkey-susanh:
#### Rationale
In Gradle 8.3, the use of `project.buildDir` was [deprecated](https://docs.gradle.org/current/userguide/upgrading_version_8.html#project_builddir) in favor of `project.layout.buildDirectory`.

#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/185

#### Changes
* Replace deprecated buildDir with layout.buildDirectory
* Add support for new property `excludeModules` that allows specifying a full path to a module (or set of modules) to exclude

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -103,11 +103,20 @@ if (hasProperty("useEmbeddedTomcat"))
 
 include BuildUtils.getMinificationProjectPath(gradle)
 
+// A list of directory names or fully qualified module paths that correspond to modules to be excluded from configuration.
+// Items may be individual module names (e.g. 'luminex'), module container names (e.g. 'customModules'), or paths (e.g., ':server:modules:commonAssays:flow')
+// You must make sure to pass this list to the appropriate calls to `BuildUtils.includeModules`
+List<String> excludedModules = []
+if (hasProperty('excludedModulePaths'))
+{
+    excludedModules.addAll("${excludedModulePaths}".split(","))
+}
+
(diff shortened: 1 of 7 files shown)


settings.gradle
```diff
@@ -103,19 +103,28 @@ if (hasProperty("useEmbeddedTomcat"))
...
 include BuildUtils.getMinificationProjectPath(gradle)
 
+// A list of directory names or fully qualified module paths that correspond to modules to be excluded from configuration.
+// Items may be individual module names (e.g. 'luminex'), module container names (e.g. 'customModules'), or paths (e.g., ':server:modules:commonAssays:flow')
+// You must make sure to pass this list to the appropriate calls to `BuildUtils.includeModules`
+List<String> excludedModules = []
+if (hasProperty('excludedModules'))
+{
+    excludedModules.addAll("${excludedModules}".split(","))
+}
```
labkey-tchad, line 113: Could the `BuildUtils.include*` methods automatically check the `excludedModules` property instead of having to check it in every settings file?

labkey-tchad (approved):

settings.gradle
labkey-susanh: I tried that, but this `hasProperty` method always returns null inside the Gradle plugin. Not sure why, but I didn't want to spend more time fighting it.

PR #624, opened 2023-11-16
Apply JavaScript syntax highlighting to `labkey:script` JSP tags

labkey-tchad:
#### Rationale
`<labkey:script>` tag bodies in JSPs are treated like plain text by default. IntelliJ can treat them like JavaScript if [configured](https://www.jetbrains.com/help/idea/using-language-injections.html#configure-injection-rules) to do so. It will do syntax highlighting and some code completion but the formatting/indentation is less than perfect.

#### Related Pull Requests
* N/A

#### Changes
* Apply JavaScript syntax highlighting to `labkey:script` JSP tags


labkey-adam (approved):
Better than before!

PR #627, opened 2023-11-28
CVE-2023-34055: Update spring version

labkey-susanh:
#### Rationale
CVE-2023-34055 - a medium-level risk but we may as well update to the latest patch version

#### Changes
* spring version update

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -103,7 +103,7 @@ apacheDirectoryVersion=2.1.3
 apacheMinaVersion=2.2.1
 
 # Keep in sync with springBootTomcatVersion below
-apacheTomcatVersion=9.0.82
+apacheTomcatVersion=9.0.83
 
 # (mothership) -> json-path -> json-smart -> accessor-smart
 # (core) -> graalvm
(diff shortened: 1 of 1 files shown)


labkey-adam:
@labkey-susanh should update `apacheTomcatVersion` as well, per the `gradle.properties` comment

labkey-susanh:
> @labkey-susanh should update `apacheTomcatVersion` as well, per the `gradle.properties` comment

Thanks. Done. If these are meant to always be the same, do we need to keep both properties still?

labkey-adam:
> > @labkey-susanh should update `apacheTomcatVersion` as well, per the `gradle.properties` comment
> 
> Thanks. Done. If these are meant to always be the same, do we need to keep both properties still?

As discussed, we could probably consolidate now or just leave as is since our embedded focus will resolve this soon. Traditionally, the two Tomcat properties were not kept in sync and we thought it important to upgrade Spring Boot and Tomcat versions in tandem. With Spring Boot 1.x EOL, we are likely to push our embedded Tomcat version past the official Spring Boot version, at least until embedded becomes universal.

labkey-adam (approved):

PR #633, opened 2023-12-01
Allow updating the CSP to allow the configured BarTender url

Sigmonia:
#### Rationale
Need to allow the BT print service URL through the CSP

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/2272
* https://github.com/LabKey/server/pull/633
* https://github.com/LabKey/platform/pull/5007

#### Changes
* Added a substitution parameter to be able to add a set of allowed `connection-src` values

### DIFF
--- a/server/bootstrap/src/org/labkey/filters/ContentSecurityPolicyFilter.java
+++ b/server/bootstrap/src/org/labkey/filters/ContentSecurityPolicyFilter.java
@@ -10,9 +10,11 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.security.SecureRandom;
+import java.util.Collection;
 import java.util.Enumeration;
-import java.util.Set;
-import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


server/bootstrap/src/org/labkey/filters/ContentSecurityPolicyFilter.java
```diff
@@ -166,4 +194,9 @@ public static String getScriptNonceHeader(HttpServletRequest request)
     }
 
     private static final SecureRandom rand = new SecureRandom();
-}
\ No newline at end of file
+
+    public static void registerAllowedConnectionSource(String allowedUrl)
```
labkey-matthewb: If we do this, I suggest something like  Map<String,Set<String>>.  Where the key is some unique string like "BartenderProvider.class"
So registerAllowedConnection(Bartender.class.getName(), Set.ofconfiguredURL)

labkey-matthewb: After any call you could do some sort of "collect" operation (remove duplicates, I think there is a stream flattener for this) and just stash the computed concatenated string.

like sets.stream().flatMap().collect(toSet)

labkey-matthewb (approved):

PR #641, opened 2023-12-11
Gradle 8.5, OWASP plugin update, update some default code styling settings for IntelliJ

labkey-susanh:
#### Rationale

- Gradle 8.5 is the latest
- OWASP 8.4.3 is the latest 8.x version. Updating to 9.x may also be possible, but locally I ran into problems, so I'll stick with the latest 8.x version for now.
- In preparation for our conversion to Tomcat 10, it is anticipated there will be fewer merge conflicts if IntelliJ doesn't collapse imports into *-imports, so this updates the style to increase the limit on when *-imports are used.  
- Some other styling changes that seem to be our standard but are not capture are also included here.

#### Related Pull Requests
* 

#### Changes
* Update gradle version
* Update OWASP dependency check plugin version
* Update IntelliJ project code style settings.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -7,10 +7,18 @@
     <option name="WHILE_ON_NEW_LINE" value="true" />
     <option name="CATCH_ON_NEW_LINE" value="true" />
     <option name="FINALLY_ON_NEW_LINE" value="true" />
+    <JSCodeStyleSettings version="0">
+      <option name="USE_DOUBLE_QUOTES" value="false" />
+      <option name="IMPORTS_WRAP" value="0" />
+      <option name="SPACES_WITHIN_OBJECT_LITERAL_BRACES" value="true" />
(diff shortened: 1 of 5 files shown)


.idea/codeStyles/Project.xml
```diff
@@ -7,10 +7,18 @@
     <option name="WHILE_ON_NEW_LINE" value="true" />
     <option name="CATCH_ON_NEW_LINE" value="true" />
     <option name="FINALLY_ON_NEW_LINE" value="true" />
+    <JSCodeStyleSettings version="0">
```
labkey-jeckels, line 10: Are these changes to settings other than Java imports intended?

labkey-jeckels (approved):

.idea/codeStyles/Project.xml
labkey-susanh: I intended them, yes, as I believe they correspond to our expected styling.

labkey-tchad (approved):

PR #650, opened 2023-12-19
HtmlView(String) inspection

labkey-matthewb:
#### Rationale
I can't delete this method yet, but error red should make these more obvious than the deprecated yellow.

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -1,6 +1,7 @@
 <component name="InspectionProjectProfileManager">
   <profile version="1.0" is_locked="false">
     <option name="myName" value="Project Default" />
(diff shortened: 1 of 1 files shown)


.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -1,6 +1,7 @@
 <component name="InspectionProjectProfileManager">
   <profile version="1.0" is_locked="false">
     <option name="myName" value="Project Default" />
+    <inspection_tool class="35f0afe5-6749-3bcd-88fe-638b65ec3b43" enabled="true" level="ERROR" enabled_by_default="true" editorAttributes="ERRORS_ATTRIBUTES" />
```
labkey-jeckels, line 4: Do we want this line?

labkey-jeckels (approved):

.idea/inspectionProfiles/Project_Default.xml
labkey-matthewb: I don't understand that line, but it's what makes this an error instead of a warning.

labkey-adam (approved):

PR #656, opened 2023-12-23
Exclude non-migrated repositories

labkey-tchad:
#### Rationale
This is more succinct now that we've migrated so many repos

#### Related Pull Requests
* 

#### Changes
*


gradle/settings/tomcat10.gradle
```diff
@@ -47,88 +47,24 @@ buildscript {
 import org.labkey.gradle.util.BuildUtils
 
 // Add modules here as they are migrated to tomcat 10
```
labkey-adam:
```suggestion
// Remove modules from here as they are migrated to tomcat 10
```

labkey-adam (approved):

PR #658, opened 2023-12-27
Prevent improper caching of `jsp2Java` task

labkey-tchad:
#### Rationale
See: https://github.com/LabKey/gradlePlugin/pull/189


labkey-susanh (approved):
Needed in tomcat10, but I assume you'll merge that in there after going to develop.

PR #672, opened 2024-01-11
Jakarta GWT

labkey-adam:
#### Rationale
Use the official Jakarta version of GWT servlet

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -62,7 +62,7 @@ windowsProteomicsBinariesVersion=1.0
 # The current version numbers for the gradle plugins.
 artifactoryPluginVersion=4.31.9
 gradleNodePluginVersion=3.5.1
-gradlePluginsVersion=2.0.0
+gradlePluginsVersion=2.1.0
 owaspDependencyCheckPluginVersion=8.4.3
 versioningPluginVersion=1.1.2
 
@@ -174,7 +174,7 @@ gsonVersion=2.8.9
 grpcVersion=1.60.1
 
 guavaVersion=33.0.0-jre
-gwtVersion=2.10.0
+gwtVersion=2.11.0
 gwtServletJakartaVersion=2.11.0
 # For dev builds, the targeted, single permutation browser. Can be either gwt-user-firefox, gwt-user-chrome, or gwt-user-ie
 gwtBrowser=gwt-user-chrome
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -175,7 +175,7 @@ grpcVersion=1.60.1
 
 guavaVersion=33.0.0-jre
 gwtVersion=2.10.0
-gwtServletVersion=2.11.0-jakarta-SNAPSHOT
+gwtServletVersion=2.11.0
```
labkey-susanh: Where is this available? I'm not seeing it on mavenCentral.

labkey-adam: https://central.sonatype.com/artifact/org.gwtproject/gwt-servlet-jakarta/versions

labkey-susanh: Ah, I see.  The coordinates are slightly different.

labkey-susanh (approved):

gradle.properties
labkey-susanh: Actually, can we update the property to gwtServletJakartaVersion to represent the new coordinates?

labkey-adam: Sure

labkey-adam: BTW: What's the deal with `gwtVersion`? Is that property use hard-coded into the plugins or something? I wasn't able to update it along with servlet, but I didn't see where it was referenced.

labkey-susanh: Yes, for better or for worse, the dependency on gwt-dev and gwt-user is set in our [Gwt gradle plugin](https://github.com/LabKey/gradlePlugin/blob/2b1db873f05ae2326585b67268b52e23ce7354dc/src/main/groovy/org/labkey/gradle/plugin/Gwt.groovy#L65).

labkey-adam: Lovely. Since you're more aware than I, can you recommend a way forward? Should I leave `gwtVersion` at `2.10.0`? Is there a new version I can set to match `gwtServletJakartaVersion`? Do we need to update the plugins?

labkey-adam: Okay, looks like it's hard-coded to look in `com.google.gwt`: https://central.sonatype.com/artifact/com.google.gwt/gwt-user/versions

But new coordinates are `org.gwtproject`: https://central.sonatype.com/artifact/org.gwtproject/gwt-user/versions

So plugins change?

labkey-adam: Old version (2.10.0) is also at the newer coordinates, so maybe we don't have to support both...

PR #679, opened 2024-01-17
Add `Reporting-Endpoints` header for embedded Tomcat

labkey-tchad:
#### Related Pull Requests
* 

#### Changes
*


labkey-tchad:
Doesn't look like we'll need to worry about this for a while

PR #680, opened 2024-01-18
Support all labkey.xml/ROOT.xml settings via application.properties

### LINKED ISSUE #5140: [LK CI] Ensure we handle all properties currently configurable through ROOT.xml/labkey.xml
[Tomcat 10 upgrade Github epic](https://github.com/LabKey/platform/issues/5140)

### Tasks 📍

- [x] Propagate JMS, LDAP, and other settings from application.properties
- [x] Update comments and error messages
- [x] Redirector for legacy context paths
- [x] Handle HTTP and HTTPS concurrently
- [x] Handle session timeout correctly
- [x] Debug config for free IntelliJ
- [x] Fix Mondrian unit tests
- [x] Fix JSP recompilation
- [x] Revisit application.properties handling for JMS, etc
- [x] Fix additional web apps📍
- [x] doc handoff

labkey-jeckels:
#### Rationale
We're enabling full compatibility for configuration via embedded Tomcat.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/5142

#### Changes
* Propagate settings from application.properties
* Update references to config files in comments

### DIFF
--- a/server/configs/application.properties
+++ b/server/configs/application.properties
@@ -21,6 +21,9 @@ context.maxWaitMillis[0]=120000
 context.accessToUnderlyingConnectionAllowed[0]=true
 context.validationQuery[0]=SELECT 1
 
+#context.logQueries[0]=true
+#context.displayName[0]=Alternate Display Name
+
 #context.dataSourceName[1]=jdbc/@@extraJdbcDataSource@@
 #context.driverClassName[1]=@@extraJdbcDriverClassName@@
 #context.url[1]=@@extraJdbcURL@@
@@ -33,6 +36,7 @@ context.encryptionKey=@@encryptionKey@@
 #context.requiredModules=
 #context.pipelineConfig=/path/to/pipeline/config/dir
 #context.serverGUID=
+#context.bypass2FA=false
 #context.workDirLocation=/path/to/desired/workDir
 
 mail.smtpHost=@@smtpHost@@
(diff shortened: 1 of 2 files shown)


server/configs/application.properties
```diff
@@ -44,6 +48,26 @@ mail.smtpUser=@@smtpUser@@
 # mail.smtpSocketFactoryClass=@@smtpSocketFactoryClass@@
 # mail.smtpAuth=@@smtpAuth@@
 
+# Optional - JMS configuration for remote ActiveMQ message management for distributed pipeline jobs
+# https://www.labkey.org/Documentation/wiki-page.view?name=jmsQueue
```
labkey-susanh, line 52: Have we captured in a TODO list somewhere that we'll need to update these docs for embedded?

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -262,10 +293,22 @@ private List<ContextResource> getDataSourceResources(ContextProperties props) th
                     dataSourceResource.setProperty("accessToUnderlyingConnectionAllowed", getPropValue(props.getAccessToUnderlyingConnectionAllowed(), i, ACCESS_TO_CONNECTION_ALLOWED_DEFAULT, "accessToUnderlyingConnectionAllowed"));
                     dataSourceResource.setProperty("validationQuery", getPropValue(props.getValidationQuery(), i, VALIDATION_QUERY_DEFAULT, "validationQuery"));
 
+                    // These two properties are handled different, as separate parameters
```
labkey-susanh:
```suggestion
                    // These two properties are handled differently, as separate parameters
```

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -283,6 +326,61 @@ private String getPropValue(Map<Integer, String> propValues, Integer resourceKey
...
+                ldapResource.setProperty("port", Integer.toString(ldapProps.getPort()));
+                if (ldapProps.getPrincipal() != null)
+                {
+                    ldapResource.setProperty("principal", ldapProps.getPrincipal());
+                }
+                if (ldapProps.getCredentials() != null)
+                {
+                    ldapResource.setProperty("credentials", ldapProps.getCredentials());
+                }
+                if (ldapProps.isUseSsl())
```
labkey-susanh: Seems like this if is not needed since we can just as well set the property to false. (Same for `isUseTls` below.)

labkey-susanh (approved):

server/embedded/src/org/labkey/embedded/LabKeyServer.java
labkey-jeckels: True. I was aiming for to create the resource most similarly to how it would have been configured in the XML world, but this simplification should be safe.

server/configs/application.properties
labkey-jeckels: Yes, there's docs handoff task in the story on the board. I'll add pointers to these two pages as definitely needing attention in the spec.

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -283,6 +326,56 @@ private String getPropValue(Map<Integer, String> propValues, Integer resourceKey
...
+                    ldapResource.setProperty("credentials", ldapProps.getCredentials());
+                }
+                ldapResource.setProperty("useSsl", Boolean.toString(ldapProps.isUseSsl()));
+                ldapResource.setProperty("useTls", Boolean.toString(ldapProps.isUseTls()));
+                if (ldapProps.getSslProtocol() != null)
+                {
+                    ldapResource.setProperty("sslProtocol", ldapProps.getSslProtocol());
+                }
+                return ldapResource;
+            }
```
labkey-adam, line 377: Thanks for taking care of this, @labkey-jeckels . I know I'm late to the party, but it seems unfortunate that `LabKeyServer` must know about every possible property used by every module. Shouldn't we introduce a registration mechanism for modules to handle the properties they care about? Similar to startup properties.

labkey-jeckels: I don't like needing centralized knowledge of these properties either, but this code has no idea about modules. And the module code doesn't know anything about application.properties or how to find it.

Some of these properties, such as disabling 2FA, could be handled via startup properties. That may be worth doing.

JMS and LDAP could probably be handled differently too but this was the approach I found that was as close to the old mechanism as possible, and easiest to manage while we're in transition between standalone and embedded.

PR #684, opened 2024-01-23
Support filtering of time fields for sql server

XingY:
#### Related Pull Requests
* https://github.com/LabKey/labkey-api-js/pull/170
* https://github.com/LabKey/labkey-ui-components/pull/1397
* https://github.com/LabKey/labkey-ui-premium/pull/309
* https://github.com/LabKey/platform/pull/5147
* https://github.com/LabKey/sampleManagement/pull/2397
* https://github.com/LabKey/server/pull/684

#### Changes
*

### DIFF
--- a/server/configs/mssql.properties
+++ b/server/configs/mssql.properties
@@ -12,7 +12,7 @@ databaseDefaultHost=localhost
 databaseDefaultPort=1433
 
 jdbcDriverClassName=com.microsoft.sqlserver.jdbc.SQLServerDriver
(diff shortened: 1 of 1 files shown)


labkey-adam:
@XingY this will work for devs, but it means that clients and our sys eng team will have to configure this property. We have a way to supply connection properties that we require. Override `prepare(LabKeyDataSource)` in the SQL Server dialect and call `dataSource.setConnectionProperty()`. Then you shouldn't need this change and no one will have to configure the new property.

XingY:
> @XingY this will work for devs, but it means that clients and our sys eng team will have to configure this property. We have a way to supply connection properties that we require. Override `prepare(LabKeyDataSource)` in the SQL Server dialect and call `dataSource.setConnectionProperty()`. Then you shouldn't need this change and no one will have to configure the new property.

Great. I'll try that out. Thanks!

PR #686, opened 2024-01-23
Update to latest Spring versions

labkey-adam:
#### Rationale
Latest is greatest


labkey-jeckels (approved):
What a nice set of changes.

PR #687, opened 2024-01-24
Support additional context paths

labkey-jeckels:
#### Rationale
Some server configurations have historically used non-root context paths, and/or deployed additional web apps for delivering static content.

#### Changes
* Property to control LabKey's context path: `context.contextPath`
* Property to automatically redirect/forward requests from a legacy context path to the root: `context.legacyContextPath`
* Properties to deploy extra web apps: `webapps.contextPath` and `webapps.docBase`

### DIFF
--- a/server/embedded/src/org/labkey/embedded/LabKeyServer.java
+++ b/server/embedded/src/org/labkey/embedded/LabKeyServer.java
@@ -193,7 +193,7 @@ public class LabKeyServer
                     {
                         if (contextProperties.getContextPath() != null && !contextProperties.getContextPath().isEmpty() && !contextProperties.getContextPath().equals("/"))
                         {
-                            throw new IllegalArgumentException("contextPath.legacyContextPath is only intended for use when deploying the LabKey application to the root context path. Please update application.properties");
+                            throw new IllegalArgumentException("contextPath.legacyContextPath is only intended for use when deploying the LabKey application to the root context path. Please update application.properties.");
                         }
                         context.addParameter("legacyContextPath", contextProperties.getLegacyContextPath());
                     }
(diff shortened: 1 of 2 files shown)


server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -183,6 +189,14 @@ protected TomcatWebServer getTomcatWebServer(Tomcat tomcat)
                         context.addParameter("OldEncryptionKey", contextProperties.getOldEncryptionKey());
                     }
 
+                    if (contextProperties.getLegacyContextPath() != null)
+                    {
+                        if (contextProperties.getContextPath() != null && !contextProperties.getContextPath().isEmpty() && !contextProperties.getContextPath().equals("/"))
+                        {
+                            throw new IllegalArgumentException("contextPath.legacyContextPath is only intended for use when deploying the LabKey application to the root context path. Please update application.properties");
```
labkey-susanh:
```suggestion
                            throw new IllegalArgumentException("contextPath.legacyContextPath is only intended for use when deploying the LabKey application to the root context path. Please update application.properties.");
```

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -156,7 +162,7 @@ protected TomcatWebServer getTomcatWebServer(Tomcat tomcat)
                     }
 
                     // set the root path to the context explicitly
-                    context.setPath("");
+                    context.setPath(contextProperties.getContextPath());
```
labkey-susanh, line 165: Any need to do trimming on this value and the legacyContextPath value? Perhaps in the setter?

labkey-susanh (approved):

server/embedded/src/org/labkey/embedded/LabKeyServer.java
labkey-jeckels: That would be fine to do, but we don't trim any of the other properties so I'm inclined to leave as-is. From a quick test, the properties file parser automatically trims leading spaces for the value (after the =) but not trailing spaces.

PR #695, opened 2024-01-29
Allow TeamCity to disable embedded tomcat

labkey-tchad:
#### Rationale
In order to make embedded tomcat the default, we need a way to turn it off for testing on TeamCity.

#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/195

#### Changes
* Update gradle plugin to get new handling of `useEmbeddedTomcat` flag

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -96,7 +96,7 @@ import org.labkey.gradle.util.BuildUtils
   node in the tree with these leaves (e.g., 'server:modules:core' will create projects for ":server"
   ":server:modules" and ":server:modules:core").
  */
-if (hasProperty("useEmbeddedTomcat") && useEmbeddedTomcat != "false")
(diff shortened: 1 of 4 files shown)


settings.gradle
```diff
@@ -96,7 +96,7 @@ import org.labkey.gradle.util.BuildUtils
   node in the tree with these leaves (e.g., 'server:modules:core' will create projects for ":server"
   ":server:modules" and ":server:modules:core").
  */
-if (hasProperty("useEmbeddedTomcat"))
+if (hasProperty("useEmbeddedTomcat") && useEmbeddedTomcat != "false")
```
labkey-susanh: I think you can use the `BuildUtils` method here.

labkey-susanh (approved):

PR #701, opened 2024-01-31
Embedded Tomcat support for custom webapp resources

labkey-tchad:
#### Rationale
A more generic solution for adding Resources to the webapp.
`labkey.xml` could have arbitrary `<Resource>` elements. This adds similar capability to `application.properties`.

This JMS resource from `labkey.xml`:
```
<Resource name="jms/ConnectionFactory" auth="Container"
        type="org.apache.activemq.ActiveMQConnectionFactory"
        factory="org.apache.activemq.jndi.JNDIReferenceFactory"
        description="JMS Connection Factory"
        brokerURL="vm://localhost?broker.persistent=false&amp;broker.useJmx=false"
        brokerName="LocalActiveMQBroker"/>
```
Can be represented like this in `application.properties`:
```
context.resources.jms.name=jms/ConnectionFactory
context.resources.jms.type=org.apache.activemq.ActiveMQConnectionFactory
context.resources.jms.factory=org.apache.activemq.jndi.JNDIReferenceFactory
context.resources.jms.description=JMS Connection Factory
context.resources.jms.brokerURL=vm://localhost?broker.persistent=false&broker.useJmx=false
context.resources.jms.brokerName=LocalActiveMQBroker
```

The `jms` in the property names is required to link related resource properties together but its exact text is unimportant.

#### Related Pull Requests
* https://github.com/LabKey/server/pull/680

#### Changes
* Add support for custom webapp resources

### DIFF
--- a/server/embedded/src/org/labkey/embedded/LabKeyServer.java
+++ b/server/embedded/src/org/labkey/embedded/LabKeyServer.java
@@ -4,6 +4,7 @@ import org.apache.catalina.core.StandardContext;
 import org.apache.catalina.loader.WebappLoader;
 import org.apache.catalina.startup.Tomcat;
 import org.apache.catalina.valves.JsonAccessLogValve;
+import org.apache.tomcat.util.collections.CaseInsensitiveKeyMap;
 import org.apache.tomcat.util.descriptor.web.ContextResource;
 import org.labkey.bootstrap.ConfigException;
 import org.springframework.boot.SpringApplication;
@@ -25,8 +26,10 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.jar.JarFile;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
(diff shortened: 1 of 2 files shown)


labkey-jeckels:
This looks great. I think it would let us rip out the `jms.` and `ldap.` examples from `application.properties` and the corresponding Java code in `LabKeyServer`, correct? The only downside is that it doesn't let us set default values for some of the properties, but we didn't have that in `labkey.xml` so that's not currently the expectation.

labkey-jeckels:
This also gives me an idea of how we can make the DataSource and extra webapp config a little cleaner...

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -246,6 +252,47 @@ private List<ContextResource> getDataSourceResources(ContextProperties props) th
...
+                List<ContextResource> contextResources = new ArrayList<>();
+                Map<String, Map<String, String>> resourceMaps = Objects.requireNonNullElse(contextProperties.getResources(), Collections.emptyMap());
+
+                for (Map.Entry<String, Map<String, String>> entry : resourceMaps.entrySet())
+                {
+                    Map<String, String> resourceMap = new CaseInsensitiveKeyMap<>();
+                    resourceMap.putAll(entry.getValue());
+                    if (!resourceMap.containsKey("name"))
+                    {
+                        logger.error("Resource configuration error: Ignoring unnamed resource 'context.resources.%s'".formatted(entry.getKey()));
```
labkey-jeckels: Some other config errors are fatal on startup, throwing an exception. I think that's probably the right direction to lean unless there's a good reason to be tolerant of an invalid config.

labkey-jeckels (approved):
See my one recommendation. After we merge the Tomcat 10 branch I'll rip out my more hard-coded handling for JMS and LDAP config.

PR #702, opened 2024-01-31
Add configuration and tasks for capturing tomcat libs for RecompilingJspClassLoader

labkey-susanh:
#### Rationale
The `RecompilingJspClassLoader` currently depends on getting the tomcat libraries it needs from the local installation of tomcat. As we march toward using embedded Tomcat, we need a different way to make these jar files available for the classpath. This PR adds a task (and its cleaning partner) to copy the same dependencies used by Gradle to a `build/jspRecompiling` directory. The task depends on the `apacheTomcatVersion` property so should get refreshed as new versions are requested.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/5174

#### Changes
* add `jspRecompileSetup` and `cleanJspRecompileSetup` tasks to the `:server` project along with the requisite configuration


labkey-jeckels (approved):
Tested locally - looks good

PR #703, opened 2024-01-31
Make 'pickPg'/'pickMssql' enable shutdown endpoint

labkey-tchad:
#### Rationale
Unless something has changed, these properties need to be set on TeamCity so that gradle can shut down Labkey.

#### Related Pull Requests
* https://github.com/LabKey/server/pull/697

#### Changes
* Make 'pickPg'/'pickMssql' enable shutdown endpoint


labkey-jeckels (approved):
I'm sorry, I didn't mean to commit the changes to `server.servlet.context-path` or `management.endpoints.web.base-path`

We can proceed with this PR as-is or restore them in the default file.

PR #713, opened 2024-02-05
Improve usability of application.properties

labkey-jeckels:
#### Rationale
`application.properties` should be easy to configure and internally consistent.

#### Changes
* Enhance `context.resources` config
* Support JDBC, LDAP, and JMS with default values and arbitrary pass-throughs
* Use property names as resource names
* Avoid using numeric indices
* Turn on JMX in embedded Tomcat scenarios

### DIFF
--- a/server/configs/application.properties
+++ b/server/configs/application.properties
@@ -14,41 +14,39 @@ server.port=@@serverPort@@
 # HTTP-only port for servers that need to handle both HTTPS (configure via server.port and server.ssl above) and HTTP
 #context.httpPort=8080
 
-context.dataSourceName[0]=jdbc/labkeyDataSource
-context.driverClassName[0]=@@jdbcDriverClassName@@
-context.url[0]=@@jdbcURL@@
-context.username[0]=@@jdbcUser@@
-context.password[0]=@@jdbcPassword@@
-context.maxTotal[0]=50
-context.maxIdle[0]=10
-context.maxWaitMillis[0]=120000
-context.accessToUnderlyingConnectionAllowed[0]=true
-context.validationQuery[0]=SELECT 1
+context.resources.jdbc.labkeyDataSource.type=javax.sql.DataSource
+context.resources.jdbc.labkeyDataSource.driverClassName=@@jdbcDriverClassName@@
+context.resources.jdbc.labkeyDataSource.url=@@jdbcURL@@
+context.resources.jdbc.labkeyDataSource.username=@@jdbcUser@@
+context.resources.jdbc.labkeyDataSource.password=@@jdbcPassword@@
+context.resources.jdbc.labkeyDataSource.maxTotal=50
+context.resources.jdbc.labkeyDataSource.maxIdle=10
+context.resources.jdbc.labkeyDataSource.maxWaitMillis=120000
+context.resources.jdbc.labkeyDataSource.accessToUnderlyingConnectionAllowed=true
+context.resources.jdbc.labkeyDataSource.validationQuery=SELECT 1
+#context.resources.jdbc.labkeyDataSource.logQueries=true
+#context.resources.jdbc.labkeyDataSource.displayName=Alternate Display Name
 
-#context.logQueries[0]=true
-#context.displayName[0]=Alternate Display Name
-
-#context.dataSourceName[1]=jdbc/@@extraJdbcDataSource@@
-#context.driverClassName[1]=@@extraJdbcDriverClassName@@
-#context.url[1]=@@extraJdbcUrl@@
-#context.username[1]=@@extraJdbcUsername@@
-#context.password[1]=@@extraJdbcPassword@@
+#context.resources.jdbc.otherDataSource.driverClassName=@@extraJdbcDriverClassName@@
+#context.resources.jdbc.otherDataSource.url=@@extraJdbcUrl@@
+#context.resources.jdbc.otherDataSource.username=@@extraJdbcUsername@@
+#context.resources.jdbc.otherDataSource.password=@@extraJdbcPassword@@
 
 #useLocalBuild#context.webAppLocation=@@pathToServer@@/build/deploy/labkeyWebapp
 context.encryptionKey=@@encryptionKey@@
 
-# By default, we'll deploy to the root context path. However, some servers have historically used /labkey or even /cpas
+# By default, we deploy to the root context path. However, some servers have historically used /labkey or even /cpas
 #context.contextPath=/labkey
 
 # Using a legacy context path provides backwards compatibility with old deployments. A typical use case would be to
-# deploy to the root context (the default) and configure /labkey as the legacy path. GETs will be redirected, and
-# non-GETs will be handled server-side via a servlet forward.
+# deploy to the root context (the default) and configure /labkey as the legacy path. GETs will be redirected.
+# All other methods (POSTs, PUTs, etc) will be handled server-side via a servlet forward.
 #context.legacyContextPath=/labkey
 
-# Other webapps to be deployed, most commonly to deliver a set of static files. Additional
-# webapps can be specified via additional indices. docBase should point at the root of the webapp's content
-#webapps.contextPath[0]=/anotherWebapp
-#webapps.docBase[0]=/my/webapp/path
+# Other webapps to be deployed, most commonly to deliver a set of static files. The context path to deploy into is the
+# property name after the "context.additionalWebapps." prefix, and the value is the location of the webapp on disk
+#context.additionalWebapps.firstContextPath=/my/webapp/path
+#context.additionalWebapps.secondContextPath=/my/other/webapp/path
 
 #context.oldEncryptionKey=
 #context.requiredModules=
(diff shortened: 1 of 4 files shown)


server/configs/application.properties
```diff
@@ -14,41 +14,39 @@ server.port=@@serverPort@@
...
-context.maxTotal[0]=50
-context.maxIdle[0]=10
-context.maxWaitMillis[0]=120000
-context.accessToUnderlyingConnectionAllowed[0]=true
-context.validationQuery[0]=SELECT 1
-
-#context.logQueries[0]=true
-#context.displayName[0]=Alternate Display Name
-
-#context.dataSourceName[1]=jdbc/@@extraJdbcDataSource@@
```
labkey-willm, line 31: Does the new style not still need a way to specify alternate data sources? You have the other 'extra' vars accounted for. below.

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -309,46 +320,33 @@ private List<ContextResource> getDataSourceResources(ContextProperties props, St
...
+                    for (Map.Entry<String, Map<String, String>> entry : parentEntry.getValue().entrySet())
                     {
-                        throw new ConfigException("Resource configuration error: Unnamed resource found 'context.resources.%s'".formatted(entry.getKey()));
-                    }
-                    if (!resourceMap.containsKey("type"))
-                    {
-                        throw new ConfigException("Resource configuration error: 'type' is not defined for resource '%s'".formatted(resourceMap.get("name")));
-                    }
+                        Map<String, String> resourceMap = new CaseInsensitiveKeyMap<>();
+                        resourceMap.putAll(entry.getValue());
```
labkey-tchad: This map copy is unused and seems unneeded.
```suggestion
```

server/configs/application.properties
labkey-jeckels: Yes, it does. I've added a comment to try to clarify how this works. Let me know if it helps or not.

server/configs/application.properties
```diff
@@ -14,41 +14,50 @@ server.port=@@serverPort@@
...
+context.resources.jdbc.labkeyDataSource.maxWaitMillis=120000
+context.resources.jdbc.labkeyDataSource.accessToUnderlyingConnectionAllowed=true
+context.resources.jdbc.labkeyDataSource.validationQuery=SELECT 1
+#context.resources.jdbc.labkeyDataSource.logQueries=true
+#context.resources.jdbc.labkeyDataSource.displayName=Alternate Display Name
+
+#context.resources.jdbc.otherDataSource.driverClassName=@@extraJdbcDriverClassName@@
+#context.resources.jdbc.otherDataSource.url=@@extraJdbcUrl@@
+#context.resources.jdbc.otherDataSource.username=@@extraJdbcUsername@@
+#context.resources.jdbc.otherDataSource.password=@@extraJdbcPassword@@
```
labkey-tchad: The extra data source name is also parameterized
```suggestion
#context.resources.jdbc.@@extraJdbcDataSource@@.driverClassName=@@extraJdbcDriverClassName@@
#context.resources.jdbc.@@extraJdbcDataSource@@.url=@@extraJdbcUrl@@
#context.resources.jdbc.@@extraJdbcDataSource@@.username=@@extraJdbcUsername@@
#context.resources.jdbc.@@extraJdbcDataSource@@.password=@@extraJdbcPassword@@
```

server/embedded/src/org/labkey/embedded/ResourceType.java
```diff
@@ -0,0 +1,113 @@
...
+            {
+                @Override
+                protected Map<String, String> mergeWithDefaults(Map<String, String> props)
+                {
+                    Map<String, String> result = super.mergeWithDefaults(props);
+                    result.putIfAbsent("type", "org.labkey.premium.ldap.LdapConnectionConfigFactory");
+                    result.putIfAbsent("factory", "org.labkey.premium.ldap.LdapConnectionConfigFactory");
+                    return result;
+                }
+            },
```
labkey-tchad: Since it is defined in the premium module, I would argue to remove the `ldap` resource type. This would also give some test coverage to the `generic` resource type by our existing ldap tests.

server/embedded/src/org/labkey/embedded/ResourceType.java
```diff
@@ -0,0 +1,113 @@
...
+    generic;
+
+    protected Map<String, String> mergeWithDefaults(Map<String, String> props)
+    {
+        Map<String, String> result = new CaseInsensitiveKeyMap<>();
+        result.putAll(props);
+        return result;
```
labkey-tchad: `mergeWithDefaults` should handle the default `auth` property, rather than my method below.
```suggestion
        result.putAll(props);
        result.putIfAbsent("auth", "Container");
        return result;
```

server/embedded/src/org/labkey/embedded/ResourceType.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        contextResource.setType(props.remove("type"));
+        contextResource.setDescription(props.remove("description"));
+        contextResource.setLookupName(props.remove("lookupName"));
+        if (props.containsKey("scope"))
+        {
+            contextResource.setScope(props.remove("scope"));
+        }
+        contextResource.setAuth(Objects.requireNonNullElse(props.remove("auth"), "Container"));
```
labkey-tchad: `mergeWithDefaults` should handle this.
```suggestion
        contextResource.setAuth(props.remove("auth"));
```

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -309,46 +320,33 @@ private List<ContextResource> getDataSourceResources(ContextProperties props, St
...
-                    contextResource.setName(resourceMap.remove("name"));
-                    contextResource.setType(resourceMap.remove("type"));
-                    contextResource.setDescription(resourceMap.remove("description"));
-                    contextResource.setLookupName(resourceMap.remove("lookupName"));
-                    if (resourceMap.containsKey("scope"))
-                    {
-                        contextResource.setScope(resourceMap.remove("scope"));
-                    }
-                    contextResource.setAuth(Objects.requireNonNullElse(resourceMap.remove("auth"), "Container"));
+                        ResourceType type;
```
labkey-tchad: Just for clarity, maybe a different variable name for this `ResourceType`. "type" is one of the resource parameters, so It's a bit confusing to have "type" mean something different here.

labkey-tchad (approved):
For consistency, should we refactor the `mail` properties in a similar way (`ResourceType.mail`)? It is also defined as a resource in `labkey.xml`.

server/configs/application.properties
```diff
@@ -66,6 +75,29 @@ mail.smtpUser=@@smtpUser@@
...
+#context.resources.jms.ConnectionFactory.brokerName=LocalActiveMQBroker
+
+# Optional - LDAP configuration for LDAP group/user synchronization
+# https://www.labkey.org/Documentation/wiki-page.view?name=LDAP_sync
+#context.resources.ldap.ConfigFactory.type=org.labkey.premium.ldap.LdapConnectionConfigFactory
+#context.resources.ldap.ConfigFactory.factory=org.labkey.premium.ldap.LdapConnectionConfigFactory
+#context.resources.ldap.ConfigFactory.host=ldap.forumsys.com
+#context.resources.ldap.ConfigFactory.port=389
+#context.resources.ldap.ConfigFactory.principal=cn=read-only-admin,dc=example,dc=com
+#context.resources.ldap.ConfigFactory.credentials=password
```
labkey-tchad: Consider using the fake ldap config from [LDAP readme](https://github.com/LabKey/premium/blob/develop/test/sampledata/ldapmock/Readme.md)
```suggestion
#context.resources.ldap.ConfigFactory.host=myldap.mydomain.com
#context.resources.ldap.ConfigFactory.port=389
#context.resources.ldap.ConfigFactory.principal=cn=read_user
#context.resources.ldap.ConfigFactory.credentials=read_user_password
```

server/embedded/src/org/labkey/embedded/ResourceType.java
labkey-jeckels: OK, done. From a quick search all configs, including Readme files, include explicit values for those two properties

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -166,11 +167,14 @@ protected TomcatWebServer getTomcatWebServer(Tomcat tomcat)
                     getDataSourceResources(contextProperties, context).forEach(contextResource -> context.getNamingResources().addResource(contextResource));
 
                     // Add extra resources to context (e.g. LDAP, JMS)
-                    getExtraContextResources(contextProperties).forEach(contextResource -> context.getNamingResources().addResource(contextResource));
+                    addExtraContextResources(contextProperties, context);
```
Sigmonia, line 169: Seems like the above getDataSourceResources call and method could use a similar treatment

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -182,7 +186,7 @@ protected TomcatWebServer getTomcatWebServer(Tomcat tomcat)
                     {
                         if (contextProperties.getContextPath() != null && !contextProperties.getContextPath().isEmpty() && !contextProperties.getContextPath().equals("/"))
                         {
-                            throw new IllegalArgumentException("contextPath.legacyContextPath is only intended for use when deploying the LabKey application to the root context path. Please update application.properties.");
+                            throw new ConfigException("contextPath.legacyContextPath is only intended for use when deploying the LabKey application to the root context path. Please update application.properties.");
```
Sigmonia, line 188: Are we catching this in the appropriate locations since this isn't a subclass of the old Exception type?

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -225,16 +229,23 @@ protected TomcatWebServer getTomcatWebServer(Tomcat tomcat)
...
+                    {
+                        String contextPath = entry.getKey();
+                        if (!contextPath.startsWith("/"))
+                        {
+                            contextPath = "/" + contextPath;
+                        }
+                        String docBase = entry.getValue();
+                        if (docBase == null || docBase.isEmpty())
+                        {
+                            throw new IllegalArgumentException("No docBase supplied additional webapp at context path " + contextPath);
```
Sigmonia: Should this be a `ConfigException`?

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -225,16 +229,23 @@ protected TomcatWebServer getTomcatWebServer(Tomcat tomcat)
...
+                        }
+                        StandardContext additionalContext = (StandardContext) tomcat.addWebapp(contextPath, docBase);
```
Sigmonia: This variable is never used

Sigmonia (approved):
Just a couple of style preferences and an unused parameter

server/embedded/src/org/labkey/embedded/LabKeyServer.java
labkey-jeckels: Updated (I debated before, but then thought it should be deleted before too long, once existing configs are migrated)

server/embedded/src/org/labkey/embedded/LabKeyServer.java
labkey-jeckels: It can now that I switched `ConfigException` to a `RuntimeException`

server/embedded/src/org/labkey/embedded/LabKeyServer.java
labkey-jeckels: Nope! These are considered fatal problems and are thrown all the way to the top, killing the startup attempt.

server/embedded/src/org/labkey/embedded/LabKeyServer.java
labkey-jeckels: These web apps aren't being deployed correctly, which I need to loop back in a future PR. I removed the variable from this FB though.

server/configs/application.properties
labkey-willm: ah, yeah, that makes sense. thanks.

PR #720, opened 2024-02-08
Fix an assortment of embedded Tomcat issues

labkey-jeckels:
#### Changes
* Avoid session deserialization outside of LK webapp
* Fix additional webapp deployments not serving files
* Support separate port for Spring Boot shutdown endpoint
* Don't use Spring Boot's error pages

### DIFF
--- a/server/embedded/src/org/labkey/embedded/LabKeyTomcatServletWebServerFactory.java
+++ b/server/embedded/src/org/labkey/embedded/LabKeyTomcatServletWebServerFactory.java
@@ -1,0 +1,517 @@
+package org.labkey.embedded;
+
+import org.apache.catalina.Host;
+import org.apache.catalina.core.StandardContext;
+import org.apache.catalina.loader.WebappLoader;
+import org.apache.catalina.startup.Tomcat;
+import org.apache.catalina.valves.JsonAccessLogValve;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.tomcat.util.descriptor.web.ContextResource;
+import org.labkey.bootstrap.ConfigException;
+import org.springframework.boot.web.embedded.tomcat.TomcatServletWebServerFactory;
+import org.springframework.boot.web.embedded.tomcat.TomcatWebServer;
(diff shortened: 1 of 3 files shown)


server/embedded/src/org/labkey/embedded/LabKeyTomcatServletWebServerFactory.java
```diff
@@ -0,0 +1,521 @@
...
+        // Prevent the Spring Boot webapp from trying to deserialize the LabKey sessions
+        getSession().setPersistent(false);
+
+        // Don't use Spring Boot's error pages, as we want to render our own
+        setErrorPages(Collections.emptySet());
+
+        super.prepareContext(host, initializers);
+    }
+
+//    @Override
```
Sigmonia: Any reason to keep this?

Sigmonia (approved):

labkey-tchad (approved):

server/embedded/src/org/labkey/embedded/LabKeyTomcatServletWebServerFactory.java
```diff
@@ -0,0 +1,521 @@
...
+                        throw new IOException("Failed to create directory " + dir + " - please check file system permissions");
+                    }
+                }
+                zipIn.closeEntry();
+                entry = zipIn.getNextEntry();
+            }
+        }
+    }
+
+    private static final int BUFFER_SIZE = 4096;
```
Sigmonia: Should probably move this up to top of the page

server/embedded/src/org/labkey/embedded/LabKeyTomcatServletWebServerFactory.java
labkey-jeckels: Nope. Deleted

server/embedded/src/org/labkey/embedded/LabKeyTomcatServletWebServerFactory.java
labkey-jeckels: I moved it. I sometimes find it helpful to keep the declaration close to the usage, but also agree that sprinkling them throughout the class isn't a good option

PR #723, opened 2024-02-14
Set pipeline config directory for embedded Tomcat

labkey-tchad:
#### Rationale
Some tests expect to put the pipeline config file in `build/deploy/embedded/config`. The server needs to be configured to look there.

#### Related Pull Requests
* https://github.com/LabKey/DiscvrLabKeyModules/pull/270

#### Changes
* Set `context.pipelineConfig` in `application.properties`

### DIFF
--- a/server/configs/application.properties
+++ b/server/configs/application.properties
@@ -61,7 +61,7 @@ context.encryptionKey=@@encryptionKey@@
 
 #context.oldEncryptionKey=
 #context.requiredModules=
-#context.pipelineConfig=/path/to/pipeline/config/dir
+context.pipelineConfig=./config
 #context.serverGUID=
 #context.bypass2FA=false
 #context.workDirLocation=/path/to/desired/workDir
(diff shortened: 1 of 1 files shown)


server/configs/application.properties
```diff
@@ -61,7 +61,7 @@ context.encryptionKey=@@encryptionKey@@
 
 #context.oldEncryptionKey=
 #context.requiredModules=
-#context.pipelineConfig=/path/to/pipeline/config/dir
+context.pipelineConfig=@@pathToServer@@/build/deploy/embedded/config
```
labkey-jeckels, line 64: Having a value for this property will cause the server to always use the "enterprise" pipeline (even if there's no JMS queue configured). I ultimately want to eliminate the non-enterprise implementation completely, but I'm not sure we're ready quite yet.

Can we easily make setting this property more conditional? If not, we can try this approach but will need to be looking more widely for pipeline problems, failure to shut down cleanly, etc.

labkey-jeckels (changes requested):

server/configs/application.properties
labkey-tchad: Sure. I can just have the task in sequenceAnalysis add this property, similar to how the premium module adds properties for ldapSync.

labkey-tchad:
Going to make this change where it matters, instead of globally

PR #728, opened 2024-02-19
Improve command-line argument for extracting resources

labkey-jeckels:
#### Rationale
`-extract` is a clearer description of what this option does

### DIFF
--- a/server/embedded/src/org/labkey/embedded/EmbeddedExtractor.java
+++ b/server/embedded/src/org/labkey/embedded/EmbeddedExtractor.java
@@ -64,10 +64,10 @@ public class EmbeddedExtractor
                     }
                 }
 
-//                if (!foundDistributionZip)
-//                {
-//                    throw new ConfigException("Unable to find distribution zip required to run LabKey Server.");
-//                }
+                if (!foundDistributionZip)
+                {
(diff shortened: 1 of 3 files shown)


labkey-jeckels:
Could also use `-extractOnly` if that sounds better.

server/embedded/src/org/labkey/embedded/EmbeddedExtractor.java
```diff
@@ -0,0 +1,149 @@
...
+                                extractFile(in, new File(pipelineLib, "servletApi.jar"));
+                            }
+                        }
+                    }
+                }
+
+//                if (!foundDistributionZip)
+//                {
+//                    throw new ConfigException("Unable to find distribution zip required to run LabKey Server.");
+//                }
```
labkey-tchad: Did you mean to make this non-fatal? I presume this is commented out for local testing?

labkey-tchad (approved):

server/embedded/src/org/labkey/embedded/EmbeddedExtractor.java
labkey-jeckels: Indeed, it was to work on my local dev build. I restored it, though we can revisit as we decide exactly how/when to extract resources in upgrade scenarios.

PR #735, opened 2024-02-21
Refactor logging configuration for embedded Tomcat

labkey-adam:
#### Rationale
We want to reconcile logging between the Spring Boot / embedded Tomcat startup process and the webapp, so both use log4j2 and our log4j2.xml for configuration. Basic approach:
* Spring Boot uses slf4j via logback, but we want to map to log4j2. So, exclude logback & log4j-to-slf4j and add log4j-slf4j2
* Point Spring Boot at our standard log4j2.xml file
* Configure log location very early in the startup process
* Add log4j2 plug-in classes to avoid startup error for unknown classes.

Questions / work to do:
* `LabKeyDeleteAction` seems to get wired up and initialized, but `execute()` is never called... neither at the Spring Boot phase nor later during webapp startup. I haven't bothered merging in the error log rotation logic because it won't ever be called.
* Not sure if `SessionAppender` works; haven't tested it yet
* Not sure if setting `logging.config` to a relative path will work everywhere

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/5257


server/bootstrap/src/org/labkey/bootstrap/LabKeyBootstrapClassLoader.java
```diff
@@ -56,10 +55,6 @@ public class LabKeyBootstrapClassLoader extends WebappClassLoader implements Exp
         String headless = "java.awt.headless";
         if (System.getProperty(headless) == null)
             System.setProperty(headless, "true");
-        // On most installs, catalina.home and catalina.base point to the same directory. However, it's possible
-        // to have multiple instances share the Tomcat binaries but have their own ./logs, ./conf, etc directories
-        // Thus, we want to use catalina.base for our place to find log files. http://www.jguru.com/faq/view.jsp?EID=1121565
-        PipelineBootstrapConfig.ensureLogHomeSet(System.getProperty("catalina.base") + "/logs");
```
labkey-jeckels, line 62: You've moved this for the embedded case, but I don't think it's being run for standalone Tomcat. I'm not aware of anyone currently deploying Tomcat and the webapp in the way described here, but seems good to retain until we're 100% embedded.

labkey-jeckels (approved):

server/bootstrap/src/org/labkey/bootstrap/LabKeyBootstrapClassLoader.java
labkey-adam: See other PR where this is addressed

PR #738, opened 2024-02-22
Apply no-op CSP filter to extra webapps

labkey-adam:
#### Rationale
Extra webapps are failing to load on standalone Tomcat when a CSP is configured in web.xml because `ContentSecurityPolicyFilter` is not accessible at filter registration time (the class was moved to the webapp). https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=49733

Simplest solution (for now, to quickly unblock the labkey.org issue) is to throw a no-op filter with the same name into bootstrap. In my local testing, this gets applied to extra webapps whereas the "real" filter gets applied to the LabKey webapp.

### DIFF
--- a/server/bootstrap/src/org/labkey/filters/ContentSecurityPolicyFilter.java
+++ b/server/bootstrap/src/org/labkey/filters/ContentSecurityPolicyFilter.java
@@ -1,0 +1,24 @@
+package org.labkey.filters;
+
+import javax.servlet.Filter;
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):
Did you test both embedded and standalone?

labkey-adam:
> Did you test both embedded and standalone?

Only standalone. Are you able to test embedded on 24.2?

PR #752, opened 2024-03-04
Improve upgrade workflow for embedded Tomcat

### LINKED ISSUE #725: [LabKey CI] Improve upgrade workflow for embedded Tomcat
[[Issue 49706](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49706)]

LabKey embedded currently has very simple logic for when to extract modules into place. If it finds the 'labkeywebapp' directory or if the 'context.webAppLocation' property is defined, it assumes that everything is already in place.
This works fine for our current primarily scenarios, test deployments and containerized deployments, because they are always setting up a fresh installation. If we want to encourage folks to switch existing production servers to embedded, the embedded server should be able to update modules in an existing deployment.

### Tasks
- [x] [Spec](https://docs.google.com/document/d/1aGvAw7QZGx38ql4Dja4H1SAIqyQZ2HB6W4CM7LRiQz8/edit) - 2
- [x] Spec review - 1 @labkey-jony
- [x] Create `application.properties` template for distributions
- [x] Remove version from 'labkeyServer.jar' included in distributions
- [x] Remove 'server' subdirectory creation
  - Update Dockerfile repo
  - Update LabKeyServer
- [x] Move VERSION file into 'labkeywebapp' directory
- [x] Compare VERSION files during startup
- [x] Add 'application.properties' to embedded distributions
- [x] Update default full-text-search index location
- [x] Backup existing deployment before extracting new distribution
- [x] Add example service scripts to embedded distributions
- [x] Code review 
  - https://github.com/LabKey/Dockerfile/pull/62
  - https://github.com/LabKey/gradlePlugin/pull/200
  - https://github.com/LabKey/testAutomation/pull/1845
  - https://github.com/LabKey/platform/pull/5295
  - https://github.com/LabKey/server/pull/752
- [x] Document upgrade process
- [x] Merge

labkey-tchad:
#### Rationale
#725 

#### Related Pull Requests
* https://github.com/LabKey/Dockerfile/pull/62
* https://github.com/LabKey/gradlePlugin/pull/200
* https://github.com/LabKey/testAutomation/pull/1845
* https://github.com/LabKey/platform/pull/5295

#### Changes
* Remove 'server' directory from embedded labkey
* Add `application.properties` for distributions
* Detect when to upgrade an existing embedded deployment


server/configs/webapps/embedded/README.txt
```diff
@@ -0,0 +1,7 @@
+Thank you for downloading LabKey Server. For more information about...
+
+- Installing LabKey Server. See https://www.labkey.org/Documentation/wiki-page.view?name=embeddedConfig
```
labkey-jeckels, line 3: @labkey-mohara what pages should we target here? Do we plan to retain the `embedded` part of these page names?

server/embedded/src/org/labkey/embedded/EmbeddedExtractor.java
```diff
@@ -16,11 +22,126 @@ public class EmbeddedExtractor
...
+                    {
+                        try (ZipInputStream zipIn = new ZipInputStream(jar.getInputStream(entry)))
+                        {
+                            ZipEntry zipEntry = zipIn.getNextEntry();
+                            // iterates over entries in the zip file
+                            while (zipEntry != null)
+                            {
+                                if (!zipEntry.isDirectory() && zipEntry.getName().equals("labkeywebapp/WEB-INF/classes/VERSION"))
+                                {
+                                    return StreamUtils.copyToString(zipIn, Charset.defaultCharset());
```
labkey-jeckels: We should use UTF-8, not whatever default charset the JVM has chosen.

server/embedded/src/org/labkey/embedded/EmbeddedExtractor.java
```diff
@@ -16,11 +22,126 @@ public class EmbeddedExtractor
...
+        {
+            throw new RuntimeException(e);
+        }
+
+        String newVersion = getNewVersion();
+
+        Version v1 = getLabKeyVersion(existingVersion);
+        Version v2 = getLabKeyVersion(newVersion);
+
+        return v1.compareTo(v2) > 0;
```
labkey-jeckels: How does this handle snapshot vs snapshot comparisons?

Additionally, I feel like we should always extract whenever the version has changed. It won't be common, but it's possible that you'd test out, say 24.3.5 and revert to 24.3.4. With this code you'd need to manually delete the extracted files first.

server/configs/webapps/embedded/config/application.properties
```diff
@@ -0,0 +1,163 @@
+server.port=8080
```
labkey-jeckels, line 1: What's the difference between this and `server/configs/application.properties`? Can we capture that in a comment in one or both files?

labkey-jeckels (changes requested):
Looking good but a few requests and questions.

server/configs/webapps/embedded/config/application.properties
labkey-tchad: Everything in `server/configs/webapps/embedded` will be included in embedded distributions. I was going to to a follow-up PR with updates to the `application.properties` and `README` files

server/configs/webapps/embedded/README.txt
labkey-tchad: For this PR, I just want to make sure these files get included. I will do a follow-up PR to update the README and application.properties template

server/embedded/src/org/labkey/embedded/EmbeddedExtractor.java
labkey-tchad: I removed the complicated version check. We now redeploy if the VERSION file has changed at all.
I've updated the VERSION file to have a link to the TeamCity build to further uniquify artifacts.
```
24.3-SNAPSHOT
https://teamcity.labkey.org/build/2941427
```
It will also check for changes to the `distribution` file so that it will redeploy if we push somebody a different distribution of the same release.

server/configs/webapps/tomcat_lk.service
```diff
@@ -0,0 +1,28 @@
+# Systemd unit file for tomcat_lk
+
+[Unit]
+Description=lk Apache Tomcat Application
```
labkey-jeckels, line 4: Not a blocker for merging, but why is `lk` in lower case and abbreviated when the rest is title caps?

labkey-jeckels (approved):

server/configs/webapps/tomcat_lk.service
labkey-tchad: These scripts are mostly just copied from Jon's samples

PR #757, opened 2024-03-05
Update to the latest OWASP plugin

labkey-adam:
#### Rationale
Docs for the latest (9.0.x) version strongly recommend obtaining and using an NVD API key, otherwise very long download times may be experienced, https://github.com/jeremylong/DependencyCheck. In my local testing, the initial download took about 35 minutes and subsequent runs took about a minute, which is inline with the previous version. We'll see how TeamCity tolerates this.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,6 +35,12 @@ allprojects {
             formats = ['HTML', 'JUNIT']
             skipConfigurations = ['dedupe', 'gwtCompileClasspath', 'gwtRuntimeClasspath', 'developmentOnly']
             skipProjects = [':server:testAutomation']
+            if (project.hasProperty('nvdApiKey'))
+            {
+                nvd {
+                    apiKey = "${project.property('nvdApiKey')}"
+                }
+            }
         }
     }
     if (BuildUtils.shouldPublish(project) || BuildUtils.shouldPublishDistribution(project))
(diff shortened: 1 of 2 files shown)


labkey-susanh:
As a second data point, the first run for me, using the NVD API key took 45 minutes for all modules and the second run does not seem like it's going to be any faster. 🤷

labkey-adam:
> As a second data point, the first run for me, using the NVD API key took 45 minutes for all modules and the second run does not seem like it's going to be any faster. 🤷

Well, better not use that NVD API key, then!

labkey-susanh (approved):

labkey-tchad (commented):
Testing on TeamCity (first try hit the suite's 15 minute time limit): https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_InternalSuites_OwaspDependencyCheck/2938105

labkey-adam:
> Testing on TeamCity (first try hit the suite's 15 minute time limit): https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_InternalSuites_OwaspDependencyCheck/2938105

Looks like second try took 36 minutes, which is consistent with my experience. @labkey-tchad what do you recommend? Should we just bump the timeout and merge this? Do we need to investigate some kind of caching strategy across agents?

labkey-tchad (approved):
36 minutes on TeamCity

PR #781, opened 2024-04-03
Compress text content by default in HTTP responses

labkey-jeckels:
#### Rationale
We used to have docs encouraging admins to enable compression in HTTP responses. With Spring Boot, we can set the default ourselves.

We don't need to adjust the size threshold is 2KB (same as our recommendations before) or MIME types from Spring Boot's defaults:

https://docs.spring.io/spring-boot/docs/current/reference/html/application-properties.html#application-properties.server.server.compression.enabled

https://www.labkey.org/Documentation/Archive/23.11/wiki-page.view?name=configTomcat#10

#### Changes
* Compress by default


labkey-adam (approved):
I'm looking forward to faster local responses

PR #793, opened 2024-04-10
embedded default cipher update

labkey-willm:
#### Rationale
The default cipher suite is outdated. This updates available cipher suites to match AWS's `ELBSecurityPolicy-TLS13-1-2-2021-06`, but is still compatible with at least `ELBSecurityPolicy-TLS-1-2-2017-01`, and maybe more. This makes it more secure than tomcat's [default set](https://github.com/apache/tomcat/blob/main/java/org/apache/tomcat/util/net/SSLHostConfig.java#L59).

#### Related Pull Requests
* https://github.com/LabKey/Dockerfile/pull/90

#### Changes
*

### DIFF
--- a/server/configs/application.properties
+++ b/server/configs/application.properties
@@ -6,14 +6,14 @@ server.port=@@serverPort@@
 ## To use ssl, update the properties below for your local installation
 
 #server.ssl.enabled=true
-#server.ssl.enabled-protocols=TLSv1.3,TLSv1.2,TLSv1.1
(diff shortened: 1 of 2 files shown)


labkey-jeckels (approved):
@labkey-mohara Unfortunately we can't set these as defaults that will automatically get applied. It should be part of the template file that we ship in new distributions, but we should specifically recommend that admins use these two properties in the docs if they're enabling HTTPS

PR #817, opened 2024-05-07
Add metric to track custom Log4J configs

labkey-jeckels:
#### Rationale
We want to be more proactive in coordinating edits to `log4j2.xml` configs for on-premise and cloud deployments

#### Changes
* Capture whether the config is custom so it can be used for a metric
* Use environment variable in another spot to simplify service configuration

### DIFF
--- a/server/configs/webapps/embedded/labkey_server.service
+++ b/server/configs/webapps/embedded/labkey_server.service
@@ -21,7 +21,6 @@ ExecStart=<JAVA_HOME>/bin/java \
     -Xms2G \
     -Xmx2G \
     -XX:+HeapDumpOnOutOfMemoryError \
-    -XX:ErrorFile=$LABKEY_HOME/logs/error_%p.log \
     -Djava.io.tmpdir=$LABKEY_HOME/labkey-tmp \
     $JAVA_FLAGS_JAR_OPS \
     $JAVA_REFLECTION_JAR_OPS \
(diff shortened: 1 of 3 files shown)


server/configs/webapps/embedded/labkey_server.service
```diff
@@ -21,7 +21,7 @@ ExecStart=<JAVA_HOME>/bin/java \
     -Xms2G \
     -Xmx2G \
     -XX:+HeapDumpOnOutOfMemoryError \
-    -XX:ErrorFile=/labkey/labkey/logs/error_%p.log \
+    -XX:ErrorFile=$LABKEY_HOME/logs/error_%p.log \
```
labkey-tchad: Any reason not to just remove the `ErrorFile` parameter entirely? You mentioned that it is unlikely to every be used and even more unlikely to be useful if it is.
Also, we don't include it in the example Windows service. I would think that the Windows and Linux examples should be functionally identical in terms of startup parameters.

labkey-tchad (approved):

server/configs/webapps/embedded/labkey_server.service
labkey-jeckels: That's fine with me. In the absence of this parameter, the JVM will write to hs_err_<pid>.log in the working directory.

labkey-susanh (approved):

PR #823, opened 2024-05-15
Use CODEOWNERS to notify people who care about Log4J config changes

labkey-jeckels:
#### Rationale
Edits to `log4j2.xml` often need to be coordinated across other systems.

#### Changes
* See if we can loop in stakeholders automatically


labkey-tchad (commented):
GitHub indicates that the SysEng team needs to be public for them to be included here

labkey-jeckels:
> GitHub indicates that the SysEng team needs to be public for them to be included here

It was public but lacked explicit write access. Now fixed.

labkey-tchad (approved):

PR #827, opened 2024-05-23
Clean up previous embedded installation before extracting upgrade

labkey-tchad:
#### Rationale
[Issue 50481: Upgrades from 24.3 to 24.4/24.5 need manual cleanup first, otherwise org.labkey.bootstrap.ConfigException](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50481)

The backup step we used to perform before attempting to self-deploy also served as a method of clearing the way for the new installation. Removing the backup step broke the upgrade process (making the upgrade require manual cleanup of the existing install).
```
org.labkey.bootstrap.ConfigException: Delete or backup existing LabKey deployment at: /usr/local/labkey/labkey/labkeywebapp
                at org.labkey.embedded.EmbeddedExtractor.extractDistributionZip(EmbeddedExtractor.java:255) ~[!/:24.5.0]
                at org.labkey.embedded.EmbeddedExtractor.extractExecutableJar(EmbeddedExtractor.java:187) ~[!/:24.5.0]
                at org.labkey.embedded.EmbeddedExtractor.extractDistribution(EmbeddedExtractor.java:165) ~[!/:24.5.0]
                at org.labkey.embedded.LabKeyTomcatServletWebServerFactory.getTomcatWebServer(LabKeyTomcatServletWebServerFactory.java:78) ~[!/:24.5.0]
```
I'm also adding validation of the contents of the `distribution.zip` to ensure we are backing up everything we should ( currently just the `modules` and `labkeywebapp` directories). This check should fail on TeamCity before a change to distribution structure could ever cause problems in the wild.

#### Related Pull Requests
* #775 

#### Changes
* Create a temporary backup while unzipping the LabKey distribution
* Add validation of distribution.zip contents (`modules` and `labkeywebapp`)

### DIFF
--- a/server/embedded/src/org/labkey/embedded/EmbeddedExtractor.java
+++ b/server/embedded/src/org/labkey/embedded/EmbeddedExtractor.java
@@ -27,6 +27,9 @@ public class EmbeddedExtractor
     private static final Log LOG = LogFactory.getLog(EmbeddedExtractor.class);
     private static final int BUFFER_SIZE = 1024 * 64;
     public static final String LABKEYWEBAPP = "labkeywebapp";
+    /**
+     * Directories that are expected to exist in 'distribution.zip'.
+     */
     private static final Set<String> EXPECTED_DIST_DIRS = Set.of(LABKEYWEBAPP, "modules");
 
     private final File currentDir = new File("").getAbsoluteFile();
--- a/server/embedded/src/org/labkey/embedded/EmbeddedExtractor.java
+++ b/server/embedded/src/org/labkey/embedded/EmbeddedExtractor.java
@@ -203,9 +203,8 @@ public class EmbeddedExtractor
         if (shouldExtract(webAppLocation))
         {
             labkeyWebappDirName = webAppLocation.getName();
-            File backupDir = createTempBackup(webAppLocation);
+            deleteOldDistribution(webAppLocation);
             extractExecutableJar(webAppLocation.getParentFile(), false);
-            deleteTempBackup(backupDir);
         }
     }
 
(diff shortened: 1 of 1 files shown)


server/embedded/src/org/labkey/embedded/EmbeddedExtractor.java
```diff
@@ -162,7 +203,9 @@ public void extractDistribution(File webAppLocation)
         if (shouldExtract(webAppLocation))
         {
             labkeyWebappDirName = webAppLocation.getName();
+            File backupDir = createTempBackup(webAppLocation);
             extractExecutableJar(webAppLocation.getParentFile(), false);
+            deleteTempBackup(backupDir);
```
labkey-jeckels: Do this in a try/finally block? It'll use a random directory name, prefixed with `labkeyBackup`, so it's possible we could leak them if there are problems extracting from the JAR

labkey-jeckels (approved):

server/embedded/src/org/labkey/embedded/EmbeddedExtractor.java
labkey-tchad: I was thinking that this way, the backup would be there if the extraction failed for some reason. That way they could, theoretically, recover to the point before attempting to upgrade. I can add some logging to help instruct folks in that direction.

labkey-jeckels: I'm curious what the recovery sequence would look like. Wouldn't they need to restore the previous JAR? And wouldn't it extract itself?

labkey-tchad: You know, that's a very good point. I'll just change this to delete the directories and not bother with the backup.

PR #843, opened 2024-07-02
add additivity=false to prevent double logging

labkey-willm:
#### Rationale
the <Root level="info"> block at the start of the Loggers could cause the subsequent calls to create double logs. Setting additivity="false" will prevent that.

https://logging.apache.org/log4j/2.x/manual/configuration.html#Additivity

#### Related Pull Requests
* https://github.com/LabKey/Dockerfile/pull/97
* https://github.com/LabKey/syseng-chef-server/pull/537

#### Changes
*

### DIFF
--- a/server/embedded/src/main/resources/log4j2.xml
+++ b/server/embedded/src/main/resources/log4j2.xml
@@ -173,19 +173,19 @@ account managers so that we can coordinate edits with other customized copies of
         </Root>
 
         <!-- category for server side script messages -->
-        <Logger name="org.labkey.api.script.ScriptService.Console" level="info">
+        <Logger name="org.labkey.api.script.ScriptService.Console" additivity="false" level="info">
             <AppenderRef ref="SessionAppender"/>
         </Logger>
 
         <!-- category for Mule messages -->
-        <Logger name="org.mule.MuleManager" level="info"/>
+        <Logger name="org.mule.MuleManager" additivity="false" level="info"/>
 
         <!--
             <Logger name="org.labkey.pipeline.mule.transformers" level="debug"/>
         -->
 
         <!-- category for jasper messages -->
-        <Logger name="org.apache.jasper" level="warn"/>
+        <Logger name="org.apache.jasper" additivity="false" level="warn"/>
 
         <Logger name="org.labkey" />
 
@@ -201,7 +201,7 @@ account managers so that we can coordinate edits with other customized copies of
             <AppenderRef ref="LABKEY"/>
         </Logger> -->
 
-        <Logger name="mondrian." level="info" />
+        <Logger name="mondrian." additivity="false" level="info" />
 
         <!--
             <Logger name="org.labkey.wiki" level="debug"/>
@@ -212,13 +212,13 @@ account managers so that we can coordinate edits with other customized copies of
         -->
 
         <!-- only show errors from PipelineJob output -->
-        <Logger name="org.labkey.api.pipeline.PipelineJob" level="error"/>
+        <Logger name="org.labkey.api.pipeline.PipelineJob" additivity="false" level="error"/>
 
         <!-- don't need to log PDFBox errors (e.g., FlateFilter, PDCIDFontType2) -->
-        <Logger name="org.apache.pdfbox" level="fatal"/>
+        <Logger name="org.apache.pdfbox" additivity="false" level="fatal"/>
 
         <!-- Suppress EHCache "maxElementsInMemory of 0" warnings, Issue 49593 -->
-        <Logger name="net.sf.ehcache.config.CacheConfiguration" level="error"/>
+        <Logger name="net.sf.ehcache.config.CacheConfiguration" additivity="false" level="error"/>
 
         <!-- this is a very verbose logger for security/permissions checking -->
         <!--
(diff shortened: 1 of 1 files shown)


labkey-adam:
I'm not an `additivity` expert by any means, but are you sure these parameters are needed here? My reading of the docs says this is useful only when a `Logger` element specifies an appender and we want its logging to go to just that appender (not the root as well). Most of these Logger elements don't specify an appender. They are present to provide the option (typically during development) of lowering the logging level for specific packages OR to raise the level for noisy third-party libraries. I don't see why we'd need `additivity` for these.

labkey-adam:
There are only three `Logger` elements here that specify an appender; I'll comment directly on those lines.

server/embedded/src/main/resources/log4j2.xml
```diff
@@ -301,7 +301,7 @@ account managers so that we can coordinate edits with other customized copies of
             <AppenderRef ref="QUERY_STATS"/>
         </Logger>
 
-        <Logger name="org.labkey.audit.event" level="OFF">
+        <Logger name="org.labkey.audit.event" additivity="false" level="OFF">
```
labkey-adam: I'm not super familiar with this logger, but clearly the intent is to log to both LABKEY_AUDIT and CONSOLE. I would remove both `additivity` and the CONSOLE ref, which should be the equivalent.

server/embedded/src/main/resources/log4j2.xml
```diff
@@ -274,11 +274,11 @@ account managers so that we can coordinate edits with other customized copies of
             <AppenderRef ref="INDEXED_DOCUMENTS"/>
         </Logger>
 
-        <Logger name="org.labkey.api.util.DebugInfoDumper" level="debug">
+        <Logger name="org.labkey.api.util.DebugInfoDumper" additivity="false" level="debug">
             <AppenderRef ref="THREAD_DUMP"/>
         </Logger>
 
-        <Logger name="org.labkey.core.admin.AdminController.ContentSecurityPolicyReportAction"  additivity="false" level="warn">
+        <Logger name="org.labkey.core.admin.AdminController.ContentSecurityPolicyReportAction" additivity="false" level="warn">
```
labkey-adam: Removing `additivity` and the CONSOLE AppenderRef should give us both console and CSP_REPORT logging.

server/embedded/src/main/resources/log4j2.xml
```diff
@@ -173,19 +173,19 @@ account managers so that we can coordinate edits with other customized copies of
         </Root>
 
         <!-- category for server side script messages -->
-        <Logger name="org.labkey.api.script.ScriptService.Console" level="info">
+        <Logger name="org.labkey.api.script.ScriptService.Console" additivity="false" level="info">
```
labkey-adam: Not sure, but I think we probably want logging to the console as well, which suggests removing additivity.

labkey-adam (changes requested):
See comments

labkey-adam:
I see now that a couple of these `additivity` parameters were already present (not added in this PR). Comments still hold, though... `additivity` plus CONSOLE appender seems unnecessary.

labkey-adam:
> I'm not an `additivity` expert by any means, but are you sure these parameters are needed here? My reading of the docs says this is useful only when a `Logger` element specifies an appender and we want its logging to go to just that appender (not the root as well). Most of these Logger elements don't specify an appender. They are present to provide the option (typically during development) of lowering the logging level for specific packages OR to raise the level for noisy third-party libraries. I don't see why we'd need `additivity` for these.

This is the very first scenario cited on the doc page as an example where `additivity` is not needed: https://logging.apache.org/log4j/2.x/manual/configuration.html#Additivity

labkey-willm:
> > I'm not an `additivity` expert by any means, but are you sure these parameters are needed here? My reading of the docs says this is useful only when a `Logger` element specifies an appender and we want its logging to go to just that appender (not the root as well). Most of these Logger elements don't specify an appender. They are present to provide the option (typically during development) of lowering the logging level for specific packages OR to raise the level for noisy third-party libraries. I don't see why we'd need `additivity` for these.
> 
> This is the very first scenario cited on the doc page as an example where `additivity` is not needed: https://logging.apache.org/log4j/2.x/manual/configuration.html#Additivity

yes, you're quite right, and this confirms what we just found with the chef version, is that having additivity=false on loggers w/o specific appenders results in no logs at all for that logger. I'll fix things up.

server/embedded/src/main/resources/log4j2.xml
```diff
@@ -293,17 +292,16 @@ account managers so that we can coordinate edits with other customized copies of
             </Logger>
          -->
 
-        <Logger name="org.labkey.core.admin.ActionsTsvWriter" additivity="false" level="info">
+        <Logger name="org.labkey.core.admin.ActionsTsvWriter" level="info">
```
labkey-adam: I think `additivity="false"` is intentional here; we want this logging to go to a separate file but not CONSOLE/LABKEY.

server/embedded/src/main/resources/log4j2.xml
```diff
@@ -293,17 +292,16 @@ account managers so that we can coordinate edits with other customized copies of
...
             <AppenderRef ref="ACTION_STATS"/>
         </Logger>
 
-        <Logger name="org.labkey.api.data.queryprofiler.QueryProfiler.QueryProfilerThread" additivity="false" level="info">
+        <Logger name="org.labkey.api.data.queryprofiler.QueryProfiler.QueryProfilerThread" level="info">
```
labkey-adam: I think `addivitity="false"` is intentional here as well

server/embedded/src/main/resources/log4j2.xml
```diff
@@ -339,6 +337,6 @@ account managers so that we can coordinate edits with other customized copies of
             <AppenderRef ref="CONSOLE"/>
         </Logger>
 
-        <Logger name="NoOpLogger" level="OFF" additivity="false"/>
+        <Logger name="NoOpLogger" level="OFF" />
```
labkey-adam: I don't know who added this or why, but guessing that `addivitity="false"` is intentional

labkey-adam (changes requested):
A couple remaining tweaks

labkey-adam (approved):

PR #860, opened 2024-08-13
Update IntelliJ run configurations and service scripts

labkey-adam:
#### Rationale
Need to add `--add-opens=java.base/java.io=ALL-UNNAMED` flag to JVM startup options for Snowflake

### DIFF
--- a/.idea/runConfigurations/LabKey_Embedded_Tomcat_Dev.xml
+++ b/.idea/runConfigurations/LabKey_Embedded_Tomcat_Dev.xml
@@ -4,7 +4,7 @@
     <option name="ALTERNATIVE_JRE_PATH_ENABLED" value="true" />
     <option name="MAIN_CLASS_NAME" value="org.labkey.embedded.LabKeyServer" />
     <module name="labkey-server.server.embedded.main" />
-    <option name="VM_PARAMETERS" value="--add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.base/java.util=ALL-UNNAMED --add-opens=java.desktop/java.awt.font=ALL-UNNAMED --add-opens=java.base/java.text=ALL-UNNAMED -Ddevmode=true -ea -Dsun.io.useCanonCaches=false -Xmx2G" />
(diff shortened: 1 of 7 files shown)


labkey-tchad (approved):
Do we know how this lists were determined? The IntelliJ run configurations don't match the embedded install scripts; nor do they match what we specify when [starting Tomcat with Gradle](https://github.com/LabKey/gradlePlugin/blob/develop/src/main/groovy/org/labkey/gradle/task/StartTomcat.groovy#L36).
They seem to match [standalone Tomcat](https://github.com/apache/tomcat/blob/main/bin/catalina.sh#L290), but not quite.

labkey-adam:
> Do we know how this lists were determined? The IntelliJ run configurations don't match the embedded install scripts; nor do they match what we specify when [starting Tomcat with Gradle](https://github.com/LabKey/gradlePlugin/blob/develop/src/main/groovy/org/labkey/gradle/task/StartTomcat.groovy#L36). They seem to match [standalone Tomcat](https://github.com/apache/tomcat/blob/main/bin/catalina.sh#L290), but not quite.

I can't tell you the full history here. These have evolved over time as we've discovered issues or added libraries that require specific flags (like the Snowflake JDBC driver). I'm not surprised that Tomcat's standard list and ours are different. Tomcat has a lot of features that LabKey never uses and LabKey has dependencies that Tomcat doesn't. We should be self consistent, though. Maybe we should try reconciling once we drop standalone Tomcat support and the corresponding startup scripts.

labkey-jeckels:
This change seems fine but I like the followup idea for making them more consistent.

@labkey-stuartm @labkey-willm @labkey-jony heads up as I imagine this would need to propagate to cloud deployments *when/if* someone attempts to adopt Snowflake. That involves adding a new data source to `application.properties`, so there would also be other config customizations for an instance to adopt it.

labkey-adam:
> This change seems fine but I like the followup idea for making them more consistent.
> 
> @labkey-stuartm @labkey-willm @labkey-jony heads up as I imagine this would need to propagate to cloud deployments _when/if_ someone attempts to adopt Snowflake. That involves adding a new data source to `application.properties`, so there would also be other config customizations for an instance to adopt it.

@labkey-stuartm @labkey-willm @labkey-jony @labkey-jeckels Snowflake usage will likely be low, and probably non-existent on our cloud deployments for the foreseeable future, but I'd still be in favor of adding this flag everywhere, including all of our startup scripts, service definitions, etc. for consistency. I've added a task to the Snowflake story.

labkey-stuartm:
> > This change seems fine but I like the followup idea for making them more consistent.
> > @labkey-stuartm @labkey-willm @labkey-jony heads up as I imagine this would need to propagate to cloud deployments _when/if_ someone attempts to adopt Snowflake. That involves adding a new data source to `application.properties`, so there would also be other config customizations for an instance to adopt it.
> 
> @labkey-stuartm @labkey-willm @labkey-jony @labkey-jeckels Snowflake usage will likely be low, and probably non-existent on our cloud deployments for the foreseeable future, but I'd still be in favor of adding this flag everywhere, including all of our startup scripts, service definitions, etc. for consistency. I've added a task to the Snowflake story.

Java_Reflection_Ops added with this PR https://github.com/FDA-MyStudies/install-script/pull/72

PR #885, opened 2024-09-20
CVE-2024-38816 Upgrade spring

labkey-martyp:
#### Rationale
We don't look to be affected by this CVE, but this will clear the OWASP tests of this particular CVE and upgrade spring and tomcat to the latest versions.

#### Changes
* Upgrade spring and tomcat versions

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -100,7 +100,7 @@ apacheDirectoryVersion=2.1.3
 apacheMinaVersion=2.2.1
 
 # Keep in sync with springBootTomcatVersion below
-apacheTomcatVersion=10.1.28
+apacheTomcatVersion=10.1.30
 
 # (mothership) -> json-path -> json-smart -> accessor-smart
 # (core) -> graalvm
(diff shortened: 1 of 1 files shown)


labkey-adam (approved):
This seems fine for suppressing the warning, but you could jump to the current versions: Tomcat 10.1.30 and SpringBoot 3.3.4.

labkey-martyp:
> This seems fine for suppressing the warning, but you could jump to the current versions: Tomcat 10.1.30 and SpringBoot 3.3.4.

Ah this is a moving target. That new spring boot version came out after I created this fb. I updated to those newer versions.

labkey-adam:
> > This seems fine for suppressing the warning, but you could jump to the current versions: Tomcat 10.1.30 and SpringBoot 3.3.4.
> 
> Ah this is a moving target. That new spring boot version came out after I created this fb. I updated to those newer versions.

They both tend to publish once a month, so just unfortunate timing.

PR #894, opened 2024-09-30
Teach EmbeddedExtractor to use distribution.properties

labkey-adam:
#### Rationale
`EmbeddedExtractor` was still using the `distribution` file

Teach EmbeddedExtractor to read `name`, `version`, and `buildUrl` properties from `distribution.properties`, if present. Logic isn't as simple as it could be (will be) since we need to continue supporting the `VERSION` file until the new gradle plugins version is adopted. The reading code currently prefers properties from `distribution.properties`, but falls back to `VERSION` if individual properties aren't found there. `VERSION` handling will be removed (and reading code simplified) once the new plugins version is adopted.

#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/220

### DIFF
--- a/server/embedded/src/org/labkey/embedded/EmbeddedExtractor.java
+++ b/server/embedded/src/org/labkey/embedded/EmbeddedExtractor.java
@@ -83,7 +83,7 @@ public class EmbeddedExtractor
 
         LabKeyDistributionInfo incomingDistribution = getDistributionInfo();
 
-        // Fresh installation
+        // Fresh installation or upgrading from a pre-distribution.properties distribution
         if (!existingVersionFile.exists() || !existingDistributionFile.exists())
         {
             LOG.info("Extracting new LabKey distribution - %s".formatted(incomingDistribution));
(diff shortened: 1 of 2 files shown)


labkey-adam:
@labkey-tchad any other testing you think we should do on this PR? As mentioned in the comment, this class will continue using the `VERSION` file until plugins v4.2.0 is published (next week?), at which point it should switch over to pulling `version` and `buildUrl` from `distribution.properties`.

server/embedded/src/org/labkey/embedded/EmbeddedExtractor.java
```diff
@@ -78,30 +79,52 @@ private File verifyJar()
...
 
         LabKeyDistributionInfo incomingDistribution = getDistributionInfo();
 
-        // Fresh installation or upgrading from non-embedded Tomcat
+        // Fresh installation
         if (!existingVersionFile.exists() || !existingDistributionFile.exists())
         {
             LOG.info("Extracting new LabKey distribution - %s".formatted(incomingDistribution));
             return true;
         }
```
labkey-tchad: This is not necessarily a fresh distribution. This could also mean we're upgrading from a pre-`distribution.properties` embedded distribution.
This will correctly trigger extraction but it won't log any information about the existing distribution. Maybe it's not worth the trouble to log more info when upgrading like that.

labkey-tchad (approved):

server/embedded/src/org/labkey/embedded/EmbeddedExtractor.java
labkey-adam: Updated comment

PR #903, opened 2024-10-09
Update test file templates

labkey-tchad:
#### Rationale
These test file templates generate files that contain warnings and errors by default.

#### Related Pull Requests
* N/A

#### Changes
* Update template user email to not cause an error when created
* Add more specific generic types to example methods and element caches

### DIFF
--- a/.idea/fileTemplates/WebDriverTestClass.java
+++ b/.idea/fileTemplates/WebDriverTestClass.java
@@ -14,7 +14,7 @@ import static org.junit.Assert.*;
 @Category({})
 public class ${NAME} extends BaseWebDriverTest
 {
-    private static final String USER = "template_user@${NAME}.test";
+    private static final String USER = "template_user@${NAME.toLowerCase()}.test";
 
     @Override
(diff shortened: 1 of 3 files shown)


.idea/fileTemplates/WebDriverTestClass.java
```diff
@@ -46,7 +46,7 @@ public void preTest()
     @Test
     public void testSomething()
     {
-        assertTrue("Test something", true);
+        assertTrue("Failing stub test", false);
```
labkey-danield, line 49: I'm assuming the change to false/fail is because it will cover more test code?

labkey-danield (approved):

labkey-chrisj (approved):

.idea/fileTemplates/WebDriverTestClass.java
labkey-tchad: This is mostly to catch when folks commit an empty template test accidentally. Now that I think about it, I should probably update the template in gradlePlugins to match these changes. That's where these stub tests usually slip in from.

PR #925, opened 2024-11-08
Update GSON and JClouds

labkey-jeckels:
#### Rationale
We can move GSON forward now that we're using the newest JClouds release

#### Changes
* Update GSON version

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -243,6 +243,10 @@ allprojects {
                     // cloud and SequenceAnalysis bring this in as a transitive dependency. We resolve to the later version here to keep things consistent
                     force "com.google.code.gson:gson:${gsonVersion}"
 
+                    // docker and cloud bring in different versions
+                    force "jakarta.ws.rs:jakarta.ws.rs-api:${jakartaWsRsApiVersion}"
+                    force "jakarta.annotation:jakarta.annotation-api:${jakartaWsAnnotationApiVersion}"
+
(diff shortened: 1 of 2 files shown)


labkey-adam:
Need to force `jakarta.ws.rs:jakarta.ws.rs-api` and `jakarta.annotation:jakarta.annotation-api`?

labkey-adam (approved):

build.gradle
```diff
@@ -331,6 +335,7 @@ allprojects {
 
                     // Force consistency for dependencies from cloud
                     force "joda-time:joda-time:${jodaTimeVersion}"
+                    force "joda-time:joda-time:${jodaTimeVersion}"
```
labkey-adam: Oops

labkey-adam (changes requested):
Looks like the "joda-time" force line got duplicated. Maybe harmless, but probably good to get rid of it.

build.gradle
labkey-jeckels: Fixed. I was hacking at it in the hopes of eliminating the force, but it's still needed (only one though!)

Sigmonia (approved):

labkey-jeckels:
Version 3.0.0 of `jakarta.ws.rs:jakarta.ws.rs-api` repackage from `javax.`->`jakarta.`. That makes it incompatible with the usage through the Docker module. Abandoning until the Docker dependency upgrades too.

PR #937, opened 2024-11-27
Slim down embedded dependencies

labkey-adam:
#### Rationale
Some dependencies in the embedded jar were included multiple times and others appear to be unnecessary. This removes 32 (~12MB) redundant, conflicting, or unneeded jars.

It's possible that one or more of these 32 jars is actually needed; @ankurjuneja may have more history here. I also plan to run BVTs and Dailies, and will merge after 24.12 branching to provide more bake time.

It's also possible we could eliminate a few more jars, but this PR seems like a good starting point.

#### Changes
* Make `tomcat-embed-core` a dependency for `bootstrap` and stop pulling in redundant Tomcat dependencies there
* Exclude `spring-boot-starter-json`: seems unused and pulls in an old version of Jackson
* Exclude `jakarta.annotation-api`: already present in `tomcat-annotations-api`
* Exclude `tomcat-juli`: already present in `tomcat-embed-core`
* Remove `spring-boot-starter-actuator`: seems unused
* Remove `spring-boot-starter-validation`: eliminating a single `@NotNull` annotation lets us remove this and its dependencies
* Migrate a single use of `org.jboss.logging.Logger` (??) to `log4j`

### DIFF
--- a/server/embedded/src/org/labkey/embedded/LabKeyServer.java
+++ b/server/embedded/src/org/labkey/embedded/LabKeyServer.java
@@ -497,6 +497,8 @@ public class LabKeyServer
 
         public String getEncryptionKey()
         {
+            if (null == encryptionKey)
+                throw new RuntimeException("Must provide encryptionKey");
             return encryptionKey;
         }
 
--- a/server/embedded/src/org/labkey/embedded/LabKeyServer.java
+++ b/server/embedded/src/org/labkey/embedded/LabKeyServer.java
@@ -1,6 +1,5 @@
 package org.labkey.embedded;
 
-import jakarta.validation.constraints.NotNull;
 import org.apache.catalina.connector.Connector;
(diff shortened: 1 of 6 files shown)


labkey-susanh (approved):

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -405,7 +404,6 @@ public static class ContextProperties
 
         private String webAppLocation;
         private String workDirLocation;
-        @NotNull (message = "Must provide encryptionKey")
```
labkey-jeckels, line 408: Fine if we want to eliminate the dependency on `NotNull` but presumably we should still enforce that the value is not null. Add a check in `getEncryptionKey()` and throw an `IllegalStateException` or similar?

labkey-adam: Done

PR #966, opened 2025-01-11
Upgrade XMLBeans to 5.3.0

labkey-adam:
#### Rationale
We've had trouble upgrading XMLBeans in the past... added logging may help track down the source of the problem

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -319,4 +319,4 @@ xercesImplVersion=2.12.2
 xmlApisVersion=1.0.b2
 
 # sync with Tika/POI
-xmlbeansVersion=5.2.0
+xmlbeansVersion=5.3.0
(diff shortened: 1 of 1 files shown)


gradle.properties
```diff
@@ -319,4 +319,4 @@ xercesImplVersion=2.12.2
 xmlApisVersion=1.0.b2
 
 # sync with Tika/POI
-xmlbeansVersion=5.2.0
+xmlbeansVersion=5.3.0
```
labkey-nicka, line 322: nit: On develop, when running gradle commands I'm seeing the XMLBeans version get logged when I don't think i was previously. Here is an example running pickPg:

image

Is this something we're logging? Is it useful? If not, could it be removed?

labkey-adam: @labkey-nicka Yes, we're logging the XMLBeans version in the Gradle Plugins and the server. As the PR comment suggests, we're trying to track down what seems like a version discrepancy between build time and run time. We'll remove the logging once we successfully upgrade.

labkey-adam:
TransformManager test case is failing an XMLBeans enum exception: https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_Bvt/3341329

PR #970, opened 2025-01-17
Minor version updates for dependencies

labkey-adam:
#### Rationale
Updates to keep current

Also, now that GraalJS is gone, remove its version constant and CVE suppression


labkey-susanh (approved):

labkey-adam:
BVTs are all green: https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_Bvt/3347580?buildTab=dependencies&type=snapshot&mode=list

PR #981, opened 2025-01-31
using `logging.log4j2.config.override` requires jackson-databind

labkey-willm:
#### Related Pull Requests
* 

#### Changes
*


labkey-willm:
this changed was added separately elsewhere

PR #956, opened 2024-12-19
Enable "enforce" CSP for dev deployments

labkey-tchad:
#### Rationale
Now that we've deployed "enforce" CSP to production servers, we should have developers run with it locally. Not enabling it by default on TeamCity because tests reliably check for CSP errors where developers might not.
The `useLocalBuild` parameter is a good proxy for non-TeamCity dev deployments.

#### Related Pull Requests
* N/A

#### Changes
* Enable the "enforce" CSP when deploying locally
* Remove example CSPs from dev `application.properties`

PR #984, opened 2025-02-06
replace CSP with new default, with start/end markers for auto replacement

labkey-willm:
#### Rationale
https://github.com/LabKey/kanban/issues/362

we're going to use the `server` repo's Content Security Policy as the source going forward. Updates there will be automatically sent to `syseng-chef-server` and `Dockerfile` repos.

Both report and enforce policies are copied in full to `syseng-chef-server`, with minor changes performed by the action. The 'enforce' policy is copied and uncommented to LabKey/Dockerfile/application.properties.

more on the action:

* Upon merge of a PR in which 'server/configs/application.properties' was updated, it copies the Content Security Policy blocks to other repos, as marked by:
* start: "## START OF CSP COPY BLOCK" or "## START OF CSP ENFORCE BLOCK" (to only copy and uncomment the csp.enforce section)
* end:  "## END OF CSP COPY BLOCK" or "## START OF CSP ENFORCE BLOCK"
* note: if the contents between the start/end have not changed, it's a no-op, and no PRs are pushed to the target repos

#### Related Pull Requests
* https://github.com/LabKey/terraform_module_ecs_service_lkapps/pull/30
* https://github.com/LabKey/syseng-chef-server/pull/608
* https://github.com/LabKey/Dockerfile/pull/135
* https://github.com/LabKey/terraform_cloud/pull/2073

#### Changes
*

### DIFF
--- a/server/configs/application.properties
+++ b/server/configs/application.properties
@@ -149,7 +149,6 @@ csp.report=\
 #useLocalBuild#    font-src 'self' data: ;\
 #useLocalBuild#    script-src 'self' 'unsafe-eval' 'strict-dynamic' 'nonce-${REQUEST.SCRIPT.NONCE}' ;\
 #useLocalBuild#    base-uri 'self' ;\
-#useLocalBuild#    upgrade-insecure-requests ;\
 #useLocalBuild#    frame-ancestors 'self' ;\
 #useLocalBuild#    report-uri /admin-contentsecuritypolicyreport.api?${CSP.REPORT.PARAMS} ;
 ## END OF CSP ENFORCE BLOCK (DO NOT CHANGE THIS TEXT)
--- a/server/configs/application.properties
+++ b/server/configs/application.properties
@@ -125,26 +125,15 @@ management.server.port=@@shutdownPort@@
 #jsonaccesslog.condition-if=attributeName
 #jsonaccesslog.condition-unless=attributeName
 
+## START OF CSP COPY BLOCK (DO NOT CHANGE THIS TEXT)
+
 ## Define one or both of 'csp.report' and 'csp.enforce' to enable Content Security Policy (CSP) headers
 ## Do not use these examples for any production environment without understanding the meaning of each directive!
 
-## Default enforce CSP for dev deployments
-#useLocalBuild#csp.enforce=\
-#useLocalBuild#    default-src 'self' https: http: ;\
-#useLocalBuild#    connect-src 'self' localhost:* ws: ${LABKEY.ALLOWED.CONNECTIONS} ;\
-#useLocalBuild#    object-src 'none' ;\
-#useLocalBuild#    style-src 'self' https: 'unsafe-inline' ;\
-#useLocalBuild#    img-src 'self' https: data: ;\
-#useLocalBuild#    font-src 'self' http: https: data: ;\
-#useLocalBuild#    script-src 'unsafe-eval' 'strict-dynamic' 'nonce-${REQUEST.SCRIPT.NONCE}' ;\
-#useLocalBuild#    base-uri 'self' ;\
-#useLocalBuild#    frame-ancestors 'self' ;\
-#useLocalBuild#    report-uri /admin-contentsecuritypolicyreport.api?${CSP.REPORT.PARAMS} ;
-
-## Default report CSP for TeamCity and dev deployments
+## Default CSP for LabKey Cloud deployments
 csp.report=\
     default-src 'self' https: http: ;\
-    connect-src 'self' localhost:* ws: ${LABKEY.ALLOWED.CONNECTIONS} ;\
+    connect-src 'self' localhost:* ${LABKEY.ALLOWED.CONNECTIONS} ;\
     object-src 'none' ;\
     style-src 'self' https: 'unsafe-inline' ;\
     img-src 'self' https: data: ;\
@@ -152,7 +141,23 @@ csp.report=\
     script-src 'unsafe-eval' 'strict-dynamic' 'nonce-${REQUEST.SCRIPT.NONCE}' ;\
     base-uri 'self' ;\
     frame-ancestors 'self' ;\
-    report-uri /admin-contentsecuritypolicyreport.api?${CSP.REPORT.PARAMS} ;
+    report-uri https://www.labkey.org/admin-contentsecuritypolicyreport.api?${CSP.REPORT.PARAMS} ;
+
+## START OF CSP ENFORCE BLOCK (DO NOT CHANGE THIS TEXT)
+# csp.enforce=\
+#    default-src 'self';\
+#    connect-src 'self' ${LABKEY.ALLOWED.CONNECTIONS} ;\
+#    object-src 'none' ;\
+#    style-src 'self' 'unsafe-inline' ;\
+#    img-src 'self' data: ;\
+#    font-src 'self' data: ;\
+#    script-src 'unsafe-eval' 'strict-dynamic' 'nonce-${REQUEST.SCRIPT.NONCE}';\
+#    base-uri 'self' ;\
+#    upgrade-insecure-requests ;\
+#    frame-ancestors 'self' ;\
+#    report-uri https://www.labkey.org/admin-contentsecuritypolicyreport.api?${CSP.REPORT.PARAMS} ;
+
+## END OF CSP COPY BLOCK (DO NOT CHANGE THIS TEXT)
 
 ## Use a custom logging configuration
 #logging.config=path/to/alternative/log4j2.xml
--- a/server/configs/application.properties
+++ b/server/configs/application.properties
@@ -125,20 +125,6 @@ management.server.port=@@shutdownPort@@
 #jsonaccesslog.condition-if=attributeName
 #jsonaccesslog.condition-unless=attributeName
 
-## START OF CSP REPORT BLOCK (DO NOT CHANGE THIS TEXT)
-csp.report=\
-    default-src 'self' ;\
-    connect-src 'self' ${LABKEY.ALLOWED.CONNECTIONS} ;\
-    object-src 'none' ;\
-    style-src 'self' 'unsafe-inline' ;\
-    img-src 'self' data: ;\
-    font-src 'self' data: ;\
-    script-src 'unsafe-eval' 'strict-dynamic' 'nonce-${REQUEST.SCRIPT.NONCE}' ;\
-    base-uri 'self' ;\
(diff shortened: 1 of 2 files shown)


labkey-tchad (changes requested):
This file is for TeamCity and dev deployments, not cloud deployments. You've removed the bit that enables the enforce CSP on dev enlistments (`#useLocalBuild#`) and made it so the CSP reports will be invisible on the test instance. TeamCity tests can't see the CSP warnings if they get sent to labkey.org.

labkey-willm:
> This file is for TeamCity and dev deployments, not cloud deployments. You've removed the bit that enables the enforce CSP on dev enlistments (`#useLocalBuild#`) and made it so the CSP reports will be invisible on the test instance. TeamCity tests can't see the CSP warnings if they get sent to labkey.org.

Ah, I see. I didn't realize what all that was doing. Do teamcity or devs need the report block at all? If not, I could leave that out of what I'm copying in from the chef repo (while retaining the #useLocalBuild# comments). 

Or better/simpler, if we _only_ copy in an uncommented enforce block, but strip "https://www.labkey.org" out of the report-uri, would that work for both dev and teamcity?

labkey-tchad:
> Ah, I see. I didn't realize what all that was doing. Do teamcity or devs need the report block at all? If not, I could leave that out of what I'm copying in from the chef repo (while retaining the #useLocalBuild# comments).
> 
> Or better/simpler, if we _only_ copy in an uncommented enforce block, but strip "https://www.labkey.org" out of the report-uri, would that work for both dev and teamcity?

Not all of our TeamCity suites can handle the `enforce` policy. RStudio still throws several warnings.
We want the `report` policy on TeamCity because the test suites will reliably check for CSP alerts in the log and alert us to problems without interrupting the rest of the test. It also lets us selectively ignore violations without disabling the CSP check entirely (https://github.com/LabKey/testAutomation/pull/1908).
For developers we want the `enforce` policy because it can make the violations more in-your-face since most folks don't monitor the csp log. (https://github.com/LabKey/server/pull/956)

labkey-willm:
> > Ah, I see. I didn't realize what all that was doing. Do teamcity or devs need the report block at all? If not, I could leave that out of what I'm copying in from the chef repo (while retaining the #useLocalBuild# comments).
> > Or better/simpler, if we _only_ copy in an uncommented enforce block, but strip "https://www.labkey.org" out of the report-uri, would that work for both dev and teamcity?
> 
> Not all of our TeamCity suites can handle the `enforce` policy. RStudio still throws several warnings. We want the `report` policy on TeamCity because the test suites will reliably check for CSP alerts in the log and alert us to problems without interrupting the rest of the test. It also lets us selectively ignore violations without disabling the CSP check entirely ([LabKey/testAutomation#1908](https://github.com/LabKey/testAutomation/pull/1908)). For developers we want the `enforce` policy because it can make the violations more in-your-face since most folks don't monitor the csp log. (#956)

ok, I've got it all fixed up to keep report enabled, enforce hidden by the `#useLocalBuild#`, the trimmed report-uri for both, and the github action in chef wired up to maintain all of that: https://github.com/LabKey/syseng-chef-server/pull/608

server/configs/application.properties
```diff
@@ -125,34 +125,34 @@ management.server.port=@@shutdownPort@@
...
+#useLocalBuild#    connect-src 'self' ${LABKEY.ALLOWED.CONNECTIONS} ;\
 #useLocalBuild#    object-src 'none' ;\
 #useLocalBuild#    style-src 'self' https: 'unsafe-inline' ;\
 #useLocalBuild#    img-src 'self' https: data: ;\
-#useLocalBuild#    font-src 'self' http: https: data: ;\
-#useLocalBuild#    script-src 'unsafe-eval' 'strict-dynamic' 'nonce-${REQUEST.SCRIPT.NONCE}' ;\
+#useLocalBuild#    font-src 'self' data: ;\
+#useLocalBuild#    script-src 'self' 'unsafe-eval' 'strict-dynamic' 'nonce-${REQUEST.SCRIPT.NONCE}' ;\
 #useLocalBuild#    base-uri 'self' ;\
+#useLocalBuild#    upgrade-insecure-requests ;\
```
labkey-tchad: Dev/Test instances generally don't have SSL enabled so upgrading insecure requests just breaks everything.

server/configs/application.properties
```diff
@@ -125,34 +125,34 @@ management.server.port=@@shutdownPort@@
 #jsonaccesslog.condition-if=attributeName
 #jsonaccesslog.condition-unless=attributeName
 
-## Define one or both of 'csp.report' and 'csp.enforce' to enable Content Security Policy (CSP) headers
-## Do not use these examples for any production environment without understanding the meaning of each directive!
+## START OF CSP REPORT BLOCK (DO NOT CHANGE THIS TEXT)
+csp.report=\
+    default-src 'self' ;\
```
labkey-tchad: Why do the `enforce` rules include `https:` but the `report` rules don't?

labkey-tchad (changes requested):

server/configs/application.properties
labkey-willm: leaving out `https:` makes the policy more strict, which needs to be vetted with `report` before it is applied to `enforce`

server/configs/application.properties
labkey-willm: I can filter that line out but would like @labkey-jeckels to weigh in first, as I recall him saying he wants to get dev/test to be using ssl just as prod does.

Would this break teamcity tests or just local dev/test things?

labkey-tchad: > Would this break teamcity tests or just local dev/test things?

Just dev. I don't think there are any suites on Teamcity that enable the `#useLocalBuild#` properties.

labkey-jeckels: > I can filter that line out but would like @labkey-jeckels to weigh in first, as I recall him saying he wants to get dev/test to be using ssl just as prod does.
> 
> Would this break teamcity tests or just local dev/test things?

Yes, I'd like to get to the point where we run HTTPS everywhere, but let's not require that on dev machines right now

server/configs/application.properties
labkey-tchad: Removing `https` causes some CSP warnings. We have some tests that embed youtube videos ([also used on labkey.org](https://www.labkey.org/SampleManagerHelp/wiki-page.view?name=folders#video)).
R reports also have the capacity to add external dependencies. JavaScript dependencies are allowed through by this CSP but CSS dependencies are not (`"violated-directive": "style-src-elem"`).

labkey-tchad (changes requested):
Not ready to remove https yet

server/configs/application.properties
labkey-willm: @labkey-adam , were any of these addressed in your recent work? 

Should I restore `https:` until these dependencies are taken care of? 

Or, proceed as is, so we can fix these and maybe catch other warnings as well.

labkey-jeckels:
> Not ready to remove https yet

Can tests and labkey.org add YouTube as an allowed external connection? I don't think we want to add that everywhere.

And should we support nonces on `style` elements too?

labkey-adam:
> > Not ready to remove https yet
> 
> Can tests and labkey.org add YouTube as an allowed external connection? I don't think we want to add that everywhere.
> 
> And should we support nonces on `style` elements too?

Yes, I was going to suggest adding `https://www.youtube.com` as well. I don't know about the `style` elements.

server/configs/application.properties
labkey-adam: I did not address these in recent work. I'd be in favor of leaving https: out of the `report` policy and adding specific youtube.com URLs as external connections.

labkey-tchad:
> And should we support nonces on `style` elements too?

The styles triggering the CSP aren't actually in a `style` element: "`<link type="text/css" rel="stylesheet" href="https://cdn.datatables.net/1.13.6/css/jquery.dataTables.min.css">`"

server/configs/application.properties
labkey-willm: ok, I've removed `upgrade-insecure-requests` from the enforce block, and added a filter to the [chef action](https://github.com/LabKey/syseng-chef-server/pull/608) to delete it before any updates get copied here.

labkey-tchad:
> > Not ready to remove https yet
> 
> Can tests and labkey.org add YouTube as an allowed external connection? I don't think we want to add that everywhere.

It would take a gradle plugin update to modify the CSP in this file when deploying dev and test environments. Probably not too tricky but not a freebie.

labkey-jeckels:
> > > Not ready to remove https yet
> > 
> > 
> > Can tests and labkey.org add YouTube as an allowed external connection? I don't think we want to add that everywhere.
> 
> It would take a gradle plugin update to modify the CSP in this file when deploying dev and test environments. Probably not too tricky but not a freebie.

I meant, can tests add allowed servers through the Admin Console's External Allowed Sources link?

labkey-tchad:
> > > > Not ready to remove https yet
> > > 
> > > 
> > > Can tests and labkey.org add YouTube as an allowed external connection? I don't think we want to add that everywhere.
> > 
> > 
> > It would take a gradle plugin update to modify the CSP in this file when deploying dev and test environments. Probably not too tricky but not a freebie.
> 
> I meant, can tests add allowed servers through the Admin Console's External Allowed Sources link?

Unfortunately, no.
`connect-src` doesn't cover iframes. We'd need a `frame-src` or `child-src` directive to allow embedding external videos.

labkey-willm:
Trey has made the excellent point that the CSP source should be in server repo, then flow on to chef/dockerfile from there. I had dismissed it early on because I thought that copy was only for local dev... didn't realize that's what all the teamcity tests were keyed from.

So now the `server` repo is the canonical source, with a github action to copy it with a few changes to the others.

labkey-tchad (approved):
Could you swap the order of the report and enforce block back to how it was? Just to make the diff a bit clearer for future history tracking. (Not a huge deal if it would have other side-effects I'm unaware of)

We'll deal with youtube embedding later. There's just one self-contained test that will trigger a CSP warning.

labkey-willm:
> Could you swap the order of the report and enforce block back to how it was? Just to make the diff a bit clearer for future history tracking. (Not a huge deal if it would have other side-effects I'm unaware of)
> 
> We'll deal with youtube embedding later. There's just one self-contained test that will trigger a CSP warning.

when I tested swapping the order back to how it was, I found and fixed a problem with the push to chef repo, in which the chef recipe version bump was happening even if the application.properties.erb was unchanged. Lucky!

PR #988, opened 2025-02-12
Add custom Log4j2 ConfigurationFactory that allows Composite Log4j configuration

Sigmonia:
#### Rationale
We need to consolidate our Log4j2.xml configs so that we can more easily update them

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/6308
* https://github.com/LabKey/Dockerfile/pull/136

#### Changes
* Add custom log4j2 ConfigurationFactory class
* update build.gradle files
* Update ClassLoaders
* Update application.properties files


labkey-willm:
does this result in 2 primary log4js that we need to keep in sync? `server/embedded/src/main/resources/configs/00.log4j2.xml` and `server/embedded/src/main/resources/log4j2.xml` ?

will these changes also impact what we're doing with chef? ( https://github.com/LabKey/syseng-chef-server/blob/develop/cookbooks/lk_appserver/templates/default/log4j2_tc10.xml.erb )

Sigmonia:
Log configuration changes moved to the Dockerfile repo.

server/embedded/src/main/resources/log4j2.xml
```diff
@@ -187,7 +187,7 @@ account managers so that we can coordinate edits with other customized copies of
         <!-- category for jasper messages -->
         <Logger name="org.apache.jasper" level="warn"/>
 
-        <Logger name="org.labkey" />
+        <Logger name="org.labkey" level="info" />
```
labkey-willm: the `<Root level="info">` above should make this (and the other two you added) moot.

labkey-willm (changes requested):
looks fine except for the added `level` bits, which shouldn't be necessary.

server/embedded/src/main/resources/log4j2.xml
Sigmonia: These were causing issues/errors when using the ConfigurationFactory, they don't seem to be doing so now. So I removed them.

labkey-willm (approved):

PR #991, opened 2025-02-17
add comments to help explain usages of the various CSP declarations

Sigmonia:
-- Note: comments are removed during application execution, so they do not appear in the Response Header

#### Rationale
Add comments to the CSP

#### Changes
* add explanatory comments to the CSP declarations

### DIFF
--- a/server/configs/application.properties
+++ b/server/configs/application.properties
@@ -125,7 +125,6 @@ management.server.port=@@shutdownPort@@
 #jsonaccesslog.condition-if=attributeName
 #jsonaccesslog.condition-unless=attributeName
 
-#### Example enforced CSP for local development
 ## START OF CSP ENFORCE BLOCK (DO NOT CHANGE THIS TEXT)
 #useLocalBuild#csp.enforce=\
 #useLocalBuild#    default-src 'self' https: ;\
--- a/server/configs/application.properties
+++ b/server/configs/application.properties
@@ -150,7 +150,7 @@ csp.report=\
     font-src 'self' data: ;                                                         /* Allowed font sources */\
     script-src 'unsafe-eval' 'strict-dynamic' 'nonce-${REQUEST.SCRIPT.NONCE}' ;     /* Allowed sources for script tags */\
     base-uri 'self' ;                                                               /* Allowed sources for base tags */\
-    frame-ancestors 'self' ;                                                        /* Allowed sources for embeded and other framed resourcees */\
+    frame-ancestors 'self' ;                                                        /* Allowed parents for embedding or framing the current resource */\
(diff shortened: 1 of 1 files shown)


server/configs/application.properties
```diff
@@ -125,6 +125,7 @@ management.server.port=@@shutdownPort@@
 #jsonaccesslog.condition-if=attributeName
 #jsonaccesslog.condition-unless=attributeName
 
+#### Example enforced CSP for local development
```
labkey-tchad: This isn't just an example. It gets deployed.

server/configs/application.properties
```diff
@@ -141,16 +142,16 @@ management.server.port=@@shutdownPort@@
...
+    connect-src 'self' ${LABKEY.ALLOWED.CONNECTIONS} ;                              /* Allowed connection sources, can be substituted and appended via the LabKey Admin UI */\
+    object-src 'none' ;                                                             /* Allowed sources for object, embed, and applet tags */\
+    style-src 'self' 'unsafe-inline' ;                                              /* Allowed sources for style tags */\
+    img-src 'self' data: ;                                                          /* Allowed sources for imgages */\
+    font-src 'self' data: ;                                                         /* Allowed font sources */\
+    script-src 'unsafe-eval' 'strict-dynamic' 'nonce-${REQUEST.SCRIPT.NONCE}' ;     /* Allowed sources for script tags */\
+    base-uri 'self' ;                                                               /* Allowed sources for base tags */\
+    frame-ancestors 'self' ;                                                        /* Allowed sources for embeded and other framed resourcees */\
```
labkey-tchad: I think this comment is describing `frame-src`. I think `frame-ancestors` is for who is allowed to embed labkey in a frame on a different site.

labkey-tchad (changes requested):
Not quite sure the point of these comments. One can look at any CSP documentation to find out what the directives mean.
It would be more useful to have comments explaining why the various directives have these settings. (e.g. why is `object-src` totally locked out)

server/configs/application.properties
labkey-willm: oops. I'd meant to remove that in my stuff.

server/configs/application.properties
Sigmonia: https://content-security-policy.com/frame-ancestors/ is where I pulled the comment from

Sigmonia: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors seems to fit with Trey's interpretation

server/configs/application.properties
Sigmonia: That gets deployed only locally or with the environment variable?

labkey-tchad: It gets deployed locally with the `useLocalBuild` gradle property. It gets deployed to production always.

labkey-tchad (approved):

labkey-willm (approved):

PR #995, opened 2025-02-19
Restore http and https for font-src in CSP for app dev hot module reloading

cnathe:
#### Rationale
The related PR made some changes to the default CSP. The change to the font-src property broke some things in the client side application hot module reloading dev experience related to how our fonts/icons are loaded.

Screenshot 2025-02-18 at 11 27 04 AM

#### Related Pull Requests
- https://github.com/LabKey/server/pull/984

#### Changes
- Restore http and https for font-src in CSP for app dev hot module reloading

### DIFF
--- a/server/configs/application.properties
+++ b/server/configs/application.properties
@@ -132,7 +132,7 @@ management.server.port=@@shutdownPort@@
 #useLocalBuild#    object-src 'none' ;\
 #useLocalBuild#    style-src 'self' https: 'unsafe-inline' ;\
 #useLocalBuild#    img-src 'self' https: data: ;\
-#useLocalBuild#    font-src 'self' data: ;\
+#useLocalBuild#    font-src 'self' http: https: data: ;\
 #useLocalBuild#    script-src 'self' 'unsafe-eval' 'strict-dynamic' 'nonce-${REQUEST.SCRIPT.NONCE}' ;\
 #useLocalBuild#    base-uri 'self' ;\
 #useLocalBuild#    frame-ancestors 'self' ;\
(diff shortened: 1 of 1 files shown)


labkey-willm:
this file is now the source for what gets deployed to prod. We can't add things back in without making corresponding changes to the "Copy CSP Blocks" action to strip them out before copying to the Dockerfile and Chef repos. I can do that but I'll wait for @labkey-adam to weigh in.

labkey-adam:
I have a local change that will allow us to add to `font-src` programmatically, just like we do `connect-src` currently. I think that's the right solution and should avoid updating the Copy action. Assuming code reviews and approvals, I should be able to merge this today.

labkey-adam:
Questions: Does this handle the `ws:` complaints as well? Can we narrow the `font-src` host to `https://localhost:3001`? `http: https:` is very broad and may allow undesired resources (on TeamCity, say).

cnathe:
@labkey-willm @labkey-adam Perfect. I was hoping that this PR would be a start to you both coming up with a better solution! Adam, I can try your suggestion of "narrow the font-src to https://localhost:3001" if you want to send me a patch.

labkey-adam:
> @labkey-willm @labkey-adam Perfect. I was hoping that this PR would be a start to you both coming up with a better solution! Adam, I can try your suggestion of "narrow the font-src to https://localhost:3001" if you want to send me a patch.

Largely untested code available in the PRs (platform and server) I just assigned. Note that you'll need to `gradlew pickpg` or similar to pick up the `application.properties` changes.

PR #996, opened 2025-02-19
Support adding module-specific resource hosts to more CSP directives

labkey-adam:
#### Rationale
Modules need the ability to allow specific font, style, frame, etc. hosts that they require

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/6345

### DIFF
--- a/server/configs/application.properties
+++ b/server/configs/application.properties
@@ -128,30 +128,30 @@ management.server.port=@@shutdownPort@@
 ## START OF CSP ENFORCE BLOCK (DO NOT CHANGE THIS TEXT)
 #useLocalBuild#csp.enforce=\
 #useLocalBuild#    default-src 'self' https: ;\
-#useLocalBuild#    connect-src 'self' ${LABKEY.ALLOWED.CONNECTION.SOURCES} ;\
+#useLocalBuild#    connect-src 'self' ${CONNECTION.SOURCES} ;\
 #useLocalBuild#    object-src 'none' ;\
-#useLocalBuild#    style-src 'self' https: 'unsafe-inline' ${LABKEY.ALLOWED.STYLE.SOURCES} ;\
+#useLocalBuild#    style-src 'self' https: 'unsafe-inline' ${STYLE.SOURCES} ;\
 #useLocalBuild#    img-src 'self' https: data: ;\
-#useLocalBuild#    font-src 'self' data: ${LABKEY.ALLOWED.FONT.SOURCES} ;\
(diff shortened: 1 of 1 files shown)


labkey-matthewb:
@labkey-willm How do we deal with branching/versions of this file.  Do we need to leave ${LABKEY.ALLOWED.CONNECTIONS} so we don't break on images with versions of LK from previous branches?

labkey-klum (approved):

labkey-willm:
> @labkey-willm How do we deal with branching/versions of this file. Do we need to leave ${LABKEY.ALLOWED.CONNECTIONS} so we don't break on images with versions of LK from previous branches?

as Stuart said in chat, "when the change is merged to the server repo, new PR's will be created in DockerFile and syseng-chef-server repos.  As long as we don't merge the syseng-chef-server change until after 25.2 is deployed, we should be ok.  We will need to merge the syseng-chef-server PR before we we do our first 25.3 deployment."

labkey-willm (approved):

server/configs/application.properties
```diff
@@ -128,28 +128,30 @@ management.server.port=@@shutdownPort@@
...
 
 ## START OF CSP REPORT BLOCK (DO NOT CHANGE THIS TEXT)
 csp.report=\
     default-src 'self' ;                                                            /* Limit the default to only the current server */\
-    connect-src 'self' ${LABKEY.ALLOWED.CONNECTIONS} ;                              /* For security purposes limit allowed connection sources, can be substituted and appended via the LabKey Admin UI */\
+    connect-src 'self' ${CONNECTION.SOURCES} ;                                      /* For security purposes limit allowed connection sources, can be substituted and appended via the LabKey Admin UI */\
     object-src 'none' ;                                                             /* These tags are not currently used by LKS */\
-    style-src 'self' 'unsafe-inline' ;                                              /* We currently have a few inline <style> tags that we are weeding out */\
+    style-src 'self' 'unsafe-inline' ${STYLE.SOURCES} ;                             /* We currently have a few inline <style> tags that we are weeding out */\
     img-src 'self' data: ;                                                          /* Limit image loading locations */\
```
labkey-jeckels, line 149: I will be surprised if we don't end up fielding a request to customize img-src at some point.

labkey-adam: Yep, probably. It will be very easy to add more directives in follow-on PRs... perhaps even the UI one coming up. I was mostly focused on getting the infrastructure in place and tested, and resolving immediate failures caused by the recent CSP changes (ReactJS hot reloading, CDS tests, etc.).

PR #1012, opened 2025-03-07
Set strong CSP: enforce for development, report-only for production

labkey-adam:
#### Rationale
For 25.3, we've decided to configure CSP r11 (report-only) by default for all on-premise deployments and CSP e11 (enforce) for all development deployments

### DIFF
--- a/server/configs/application.properties
+++ b/server/configs/application.properties
@@ -124,17 +124,17 @@ management.server.port=@@shutdownPort@@
 
 ## START OF CSP ENFORCE BLOCK (DO NOT CHANGE THIS TEXT)
 #useLocalBuild#csp.enforce=\
-#useLocalBuild#    default-src 'self' https: ; \
-#useLocalBuild#    connect-src 'self' ${CONNECTION.SOURCES} ; \
-#useLocalBuild#    object-src 'none' ; \
-#useLocalBuild#    style-src 'self' https: 'unsafe-inline' ${STYLE.SOURCES} ; \
-#useLocalBuild#    img-src 'self' https: data: ${IMAGE.SOURCES} ; \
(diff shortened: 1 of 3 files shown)


labkey-jeckels:
This is unlikely to actually impact existing installs. Should we instead or also make this the default in `LabKeyServer.java`, which can then be overridden as admins desire?

labkey-adam:
> This is unlikely to actually impact existing installs. Should we instead or also make this the default in `LabKeyServer.java`, which can then be overridden as admins desire?

I don't actually know how these properties get applied on production servers. `application.properties` is ignored?

labkey-adam:
> > This is unlikely to actually impact existing installs. Should we instead or also make this the default in `LabKeyServer.java`, which can then be overridden as admins desire?
> 
> I don't actually know how these properties get applied on production servers. `application.properties` is ignored?

Oh, *existing* installs... because they don't replace application.properties every upgrade?

labkey-jeckels:
> > > This is unlikely to actually impact existing installs. Should we instead or also make this the default in `LabKeyServer.java`, which can then be overridden as admins desire?
> > 
> > 
> > I don't actually know how these properties get applied on production servers. `application.properties` is ignored?
> 
> Oh, _existing_ installs... because they don't replace application.properties every upgrade?

Exactly. It has their DB settings and other install-specific config. Ideally they might check the new version in the download and merge them together, but I doubt that happens very often

labkey-adam:
@labkey-jeckels please review the latest. I didn't see much value in keeping the comments in the code version.

labkey-willm (approved):

labkey-jeckels (approved):
It's unfortunate to have another copy of CSP checked in. Maybe we can soon have both a report and enforce specified solely in `LabKeyServer.java`. We'd need to conditionalize the `report-uri`

PR #1033, opened 2025-04-02
fix exception when loading log4j2

Sigmonia:
#### Rationale
[Issue 52671](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52671): TeamCity logs are complaining about the logging configuration

#### Changes
* adjust level declaration


Sigmonia:
changing to target 25.3

PR #1050, opened 2025-04-22
support testing with serverContext on teamCity

labkey-chrisj:
#### Rationale
This change accompanies the changes made in https://github.com/LabKey/gradlePlugin/pull/231 to support configuring teamCity runs to use a contextPath
Github story: https://github.com/LabKey/kanban/issues/664

#### Related Pull Requests
updates to gradle plug-in: https://github.com/LabKey/gradlePlugin/pull/231

#### Changes
updates the gradle plug-in version

### DIFF
--- a/server/configs/webapps/embedded/config/application.properties
+++ b/server/configs/webapps/embedded/config/application.properties
@@ -67,7 +67,7 @@ server.error.include-message=always
 
 ## We do not recommend deploying to a non-root context path (e.g. http://localhost:8080/labkey)
 ## If you do set this property, do not set a legacyContextPath.
-#context.contextPath=/@@contextPath@@
+#context.contextPath=@@contextPath@@
 
 ####################################################################################
 ## SMTP configuration
--- a/server/configs/webapps/embedded/config/application.properties
+++ b/server/configs/webapps/embedded/config/application.properties
@@ -67,7 +67,7 @@ server.error.include-message=always
 
 ## We do not recommend deploying to a non-root context path (e.g. http://localhost:8080/labkey)
 ## If you do set this property, do not set a legacyContextPath.
-#context.contextPath=/labkey
+#context.contextPath=/@@contextPath@@
 
 ####################################################################################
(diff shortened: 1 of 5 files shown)


labkey-tchad (commented):
You downgraded a bunch of dependencies somehow. don't do that

labkey-chrisj:
> You downgraded a bunch of dependencies somehow. don't do that

Yikes. Pushed a change to align with right-now develop

server/configs/webapps/embedded/config/application.properties
```diff
@@ -67,7 +67,7 @@ server.error.include-message=always
 
 ## We do not recommend deploying to a non-root context path (e.g. http://localhost:8080/labkey)
 ## If you do set this property, do not set a legacyContextPath.
-#context.contextPath=/labkey
+#context.contextPath=@@contextPath@@
```
labkey-tchad: This file (`server/configs/webapps/embedded/config/application.properties`) is the one we bundle with distributions. TeamCity doesn't care about it. Changing this _probably_ won't cause any trouble but folks might have scripts that depend on the current text. Let's leave it as is; not worth the risk IMO.

labkey-tchad (approved):

server/configs/webapps/embedded/config/application.properties
labkey-chrisj: If that change isn't there, I worry the plugin won't uncomment that line.  It does this check: 
(DoThenSetup.groovy line 110)
`if (configProperties.containsKey("contextPath") && line.contains("=@@contextPath"))
                    {
                        line = line.replace("#context.", "context.")
                    }`

If I undo this change, I think I'll have to check differently or else we don't effectively configure the server for the run.
...as I review my changes and the results, it's the one at `server/configs/application.properties` that seems to have resolved the problem of that line not being uncommented.  I'll try reverting that change in the one at `server/configs/webapps/embedded/config/` to see if it still works

(edit: it does work.)

PR #1056, opened 2025-04-30
bump springboot etc for tomcat CVE-2025-31650

labkey-willm:
#### Rationale
https://ossindex.sonatype.org/vulnerability/CVE-2025-31650

same as in https://github.com/LabKey/server/pull/1055

#### Related Pull Requests
* 

#### Changes
*


labkey-adam (approved):
This PR is not strictly required since we'll merge 25.3 -> develop soon enough, but shouldn't hurt

PR #1057, opened 2025-04-30
Exclude some more places that offer NotNull and Nullable

labkey-susanh:
#### Rationale
It seems that everyone wants to produce `NotNull` and `Nullable` classes these days. We aren't really interested in the others.

#### Changes
* Update `codeInsightSettings.xml` with more exclusions


labkey-adam (approved):
Thanks for this!

PR #1065, opened 2025-05-14
Enable a strong enforce CSP by default

labkey-adam:
#### Rationale
Set a strong enforce CSP by default. Remove CSP properties from `application.properties` files.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/6670
* https://github.com/LabKey/premiumModules/pull/258


labkey-matthewb (approved):

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -76,27 +76,35 @@ public static void main(String[] args)
                 font-src 'self' data: ${FONT.SOURCES} ;
                 script-src 'unsafe-eval' 'strict-dynamic' 'nonce-${REQUEST.SCRIPT.NONCE}' ;
                 base-uri 'self' ;
+                frame-src 'self' ${FRAME.SOURCES} ;
+            """;
+        // Add upgrade_insecure_requests substitution, frame-ancestors, and e12 version for enforce CSP
+        String enforceCsp = baseCsp + """
                 ${UPGRADE.INSECURE.REQUESTS}
```
labkey-willm, line 83: does this not need a ; at the end?

labkey-adam: The substitution adds the ;

This way, if running without HTTPS we don't get an extra semicolon

labkey-adam: @labkey-willm https://github.com/LabKey/platform/blob/develop/api/src/org/labkey/filters/ContentSecurityPolicyFilter.java#L300

PR #1074, opened 2025-05-28
Suppress dependency check false-positive for mxparser

labkey-tchad:
#### Rationale
The OWASP checker started matching numerous XStream CVEs to mxparser. None of the CVEs are actually associated with mxparser.

#### Related Pull Requests
* N/A

#### Changes
* Suppress OWASP false-positive for mxparser


labkey-jeckels (approved):
Agreed that this is a bad match/false positive.

PR #1083, opened 2025-06-02
Issue 53209: Allow externalModules folder to be dictated via application.properties file

labkey-jeckels:
#### Rationale
We've long supported pointing to a custom external modules directory via a system property. We can also support using `application.properties`

#### Changes
* Look for `context.externalModules` in `application.properties` and use it to set the system property

### DIFF
--- a/server/configs/application.properties
+++ b/server/configs/application.properties
@@ -58,7 +58,7 @@ context.encryptionKey=@@encryptionKey@@
 #context.additionalWebapps.firstContextPath=/my/webapp/path
 #context.additionalWebapps.secondContextPath=/my/other/webapp/path
 
-context.externalModules=/Users/jeckels/externalModules
+#context.externalModules=/path/to/external/modules/dir
 #context.requiredModules=
 #context.pipelineConfig=/path/to/pipeline/config/dir
 #context.serverGUID=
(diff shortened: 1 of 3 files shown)


labkey-nicka (approved):

labkey-jeckels:
@labkey-jony installers queued for this feature branch: https://teamcity.labkey.org/buildConfiguration/LabKey_253Release_Premium_InstallersCustom/3534685

labkey-jony:
Confirmed working.

Performed the following test:

- Spun up a trial instance and stopped Tomcat.
- Downloaded the labkey.org binaries mentioned above and installed the labkeyServer.jar to `/labkey/labkey`, replacing the original one for trial.
- Copied the evaluationContent.module file from `/labkey/labkey/modules` to `/labkey/labkey/externalModules` (I needed a module that didn't exist in the labkey.org binaries)
- Sanitized the trial instance for the test (removed startup properties files, dropped and recreated the database so it was clean, etc).
- Started Tomcat and went through the initial setup wizard
- Checked the module list on the Admin Console and confirmed that evaluationContent was there and mapped to `/labkey/labkey/externalModules`
- Stopped Tomcat and created a new optional.application.properties file
- Added the following line to optional.application.properties:

`context.externalModules=/labkey/testdirectory/externalModules`

- Created the folders to make `/labkey/testdirectory/externalModules` and moved the evaluationContent.module to that directory
- Started Tomcat and rechecked the modules page on the Admin Console
- Confirmed the module was still there and was mapping to the new location when looking at the module details.

labkey-jony (approved):

PR #1087, opened 2025-06-09
Re-enable shutdown endpoint

labkey-tchad:
#### Rationale
[Issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53244)
The shutdown endpoint was inadvertently disabled when the `spring-boot-starter-actuator` dependency was removed in 25.1. The `stopLabkey` Gradle task requires this endpoint.

#### Related Pull Requests
* https://github.com/LabKey/server/pull/937

#### Changes
* Add `spring-boot-starter-actuator` dependency


labkey-tchad:
Decided to drop support for the shutdown endpoint. Just need to remove the unnecessary properties and the dead code in our gradle plugin

PR #1092, opened 2025-06-12
Update to Postgres driver version 42.7.7

labkey-susanh:
#### Rationale
[CVE-2025-49146](https://ossindex.sonatype.org/vulnerability/CVE-2025-49146?component-type=maven&component-name=org.postgresql%2Fpostgresql&utm_source=dependency-check&utm_medium=integration&utm_content=12.1.1) Affects versions from 42.7.4 until 42.7.7. We are not at risk since we do not configure channel binding, but good to update just the same.

#### Changes
* Update `postgresDriverVersion`

PR #1093, opened 2025-06-13
Suppress CVE-2025-49146

labkey-susanh:
#### Rationale
Turns out the latest version of the postgres driver has some significant performance issues, so instead of updating, we'll suppress [CVE-2025-49146](https://ossindex.sonatype.org/vulnerability/CVE-2025-49146?component-type=maven&component-name=org.postgresql%2Fpostgresql&utm_source=dependency-check&utm_medium=integration&utm_content=12.1.1) since we are not at risk from it

#### Related Pull Requests
* #1092

#### Changes
* Add suppression

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -266,6 +266,7 @@ poiVersion=5.4.0
 
 pollingWatchVersion=0.2.0
 
+# Newer versions of the driver have a perf degradation that's important for us. https://github.com/pgjdbc/pgjdbc/issues/3505
 postgresqlDriverVersion=42.7.4
 
 quartzVersion=2.5.0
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -266,7 +266,7 @@ poiVersion=5.4.0
 
 pollingWatchVersion=0.2.0
 
-postgresqlDriverVersion=42.7.7
+postgresqlDriverVersion=42.7.4
```
labkey-jeckels, line 270:
```suggestion
# Newer versions of the driver have a perf degradation that's important for us. https://github.com/pgjdbc/pgjdbc/issues/3505
postgresqlDriverVersion=42.7.4
```

labkey-jeckels (approved):
A comment here seems like a helpful complement to the note in the doc

gradle.properties
labkey-susanh: Good idea. done.

PR #1101, opened 2025-06-19
CSP reports don't get reported locally if using a contextPath

labkey-adam:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53319

Adding the context path to the `report-uri` does the trick -- tested both with and without a context path. Note: for this to work correctly, any provided `contextPath` value must start with a slash. Credit to @labkey-tchad for current approach.

#### Related Pull Requests
- https://github.com/LabKey/server/pull/1103

### DIFF
--- a/server/configs/application.properties
+++ b/server/configs/application.properties
@@ -140,7 +140,7 @@ management.server.port=@@shutdownPort@@
 #useLocalBuild#    base-uri 'self' ;                                                                       /* Limit the base tags to only source from current server */\
 #useLocalBuild#    frame-ancestors 'self' ;                                                                /* Limit iframe content destinations (who can load this server's content into an iframe) */\
 #useLocalBuild#    frame-src 'self' ${FRAME.SOURCES} ;                                                     /* Limit iframe content sources (from what servers can this server's iframe content be loaded) */\
-#useLocalBuild#    report-uri /admin-contentSecurityPolicyReport.api?cspVersion=e11&${CSP.REPORT.PARAMS} ; /* Report any encountered CSP violations to the supplied URL */
+#useLocalBuild#    report-uri admin-contentSecurityPolicyReport.api?cspVersion=e11&${CSP.REPORT.PARAMS} ; /* Report any encountered CSP violations to the supplied URL */
 ## END OF CSP ENFORCE BLOCK (DO NOT CHANGE THIS TEXT)
 
 ## START OF CSP REPORT BLOCK (DO NOT CHANGE THIS TEXT)
@@ -155,7 +155,7 @@ csp.report=\
     base-uri 'self' ;                                                                       /* Limit the base tags to only source from current server */\
     frame-ancestors 'self' ;                                                                /* Limit iframe content destinations (who can load this server's content into an iframe) */\
     frame-src 'self' ${FRAME.SOURCES} ;                                                     /* Limit iframe content sources (from what servers can this server's iframe content be loaded) */\
-    report-uri /admin-contentSecurityPolicyReport.api?cspVersion=r11&${CSP.REPORT.PARAMS} ; /* Report any encountered CSP violations to the supplied URL */
+    report-uri admin-contentSecurityPolicyReport.api?cspVersion=r11&${CSP.REPORT.PARAMS} ; /* Report any encountered CSP violations to the supplied URL */
 ## END OF CSP REPORT BLOCK (DO NOT CHANGE THIS TEXT)
 
 ## Use a custom logging configuration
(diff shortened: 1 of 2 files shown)


labkey-matthewb (approved):

labkey-matthewb (changes requested):
On second thought, this might not be right.  This CSP will also be used for static pages.  In that case this might be a nonsense URL.  E.g. looks like an ActionURL with an improper container path.

labkey-tchad (commented):
The CSPs in `application.properties` are only used for dev enlistments, which use a root contextPath without some modification to the file. If they do that, I think it's reasonable to expect them to modify the CSP as well.

The CSP defined in Java has access to the contextPath and could just include it in the CSP's `report-uri`, right?

server/configs/application.properties
```diff
@@ -140,7 +140,7 @@ management.server.port=@@shutdownPort@@
 #useLocalBuild#    base-uri 'self' ;                                                                       /* Limit the base tags to only source from current server */\
 #useLocalBuild#    frame-ancestors 'self' ;                                                                /* Limit iframe content destinations (who can load this server's content into an iframe) */\
 #useLocalBuild#    frame-src 'self' ${FRAME.SOURCES} ;                                                     /* Limit iframe content sources (from what servers can this server's iframe content be loaded) */\
-#useLocalBuild#    report-uri /admin-contentSecurityPolicyReport.api?cspVersion=e11&${CSP.REPORT.PARAMS} ; /* Report any encountered CSP violations to the supplied URL */
+#useLocalBuild#    report-uri admin-contentSecurityPolicyReport.api?cspVersion=e11&${CSP.REPORT.PARAMS} ;  /* Report any encountered CSP violations to the supplied URL */
```
labkey-tchad:
```suggestion
#useLocalBuild#    report-uri ${context.contextPath:}/admin-contentSecurityPolicyReport.api?cspVersion=e11&${CSP.REPORT.PARAMS} ;  /* Report any encountered CSP violations to the supplied URL */
```

server/configs/application.properties
```diff
@@ -155,7 +155,7 @@ csp.report=\
     base-uri 'self' ;                                                                       /* Limit the base tags to only source from current server */\
     frame-ancestors 'self' ;                                                                /* Limit iframe content destinations (who can load this server's content into an iframe) */\
     frame-src 'self' ${FRAME.SOURCES} ;                                                     /* Limit iframe content sources (from what servers can this server's iframe content be loaded) */\
-    report-uri /admin-contentSecurityPolicyReport.api?cspVersion=r11&${CSP.REPORT.PARAMS} ; /* Report any encountered CSP violations to the supplied URL */
+    report-uri admin-contentSecurityPolicyReport.api?cspVersion=r11&${CSP.REPORT.PARAMS} ;  /* Report any encountered CSP violations to the supplied URL */
```
labkey-tchad:
```suggestion
    report-uri ${context.contextPath:}/admin-contentSecurityPolicyReport.api?cspVersion=r11&${CSP.REPORT.PARAMS} ;  /* Report any encountered CSP violations to the supplied URL */
```

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -94,7 +94,7 @@ public static void main(String[] args)
                         base-uri 'self' ;
                         frame-ancestors 'self' ;
                         frame-src 'self' ${FRAME.SOURCES} ;
-                        report-uri /admin-contentSecurityPolicyReport.api?cspVersion=r11&${CSP.REPORT.PARAMS}
+                        report-uri admin-contentSecurityPolicyReport.api?cspVersion=r11&${CSP.REPORT.PARAMS}
```
labkey-tchad:
```suggestion
                        report-uri ${context.contextPath:}/admin-contentSecurityPolicyReport.api?cspVersion=r11&${CSP.REPORT.PARAMS}
```

labkey-tchad (changes requested):
Spring boot support property interpolation. If we prepend the report URI with `${context.contextPath:}`, it will fill in the context path (if it is defined)

labkey-adam:
> The CSPs in `application.properties` are only used for dev enlistments, which use a root contextPath without some modification to the file. If they do that, I think it's reasonable to expect them to modify the CSP as well.
> 
> The CSP defined in Java has access to the contextPath and could just include it in the CSP's `report-uri`, right?

Given that these are dev-only and short lived (removed starting in 25.6), and the fact that @labkey-willm automation relies on this URI (I think), I'm going to leave the `application.properties` URIs as is.

server/embedded/src/org/labkey/embedded/LabKeyServer.java
labkey-adam: I like this, though I think this works only if they specify a leading forward slash in their contextPath property. I think the property is respected with or without that slash.

labkey-matthewb (approved):

labkey-tchad (approved):

labkey-adam:
> On second thought, this might not be right. This CSP will also be used for static pages. In that case this might be a nonsense URL. E.g. looks like an ActionURL with an improper container path.

Approach was changed to continue to use an absolute URI but explicitly include the context path

PR #1111, opened 2025-06-27
Remove application properties for configuring management endpoints

labkey-tchad:
#### Rationale
[Issue 53244: Unable to stop server via Spring Boot shutdown endpoint](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53244)

We've decided it's ok to stop supporting this functionality.

#### Related Pull Requests
* https://github.com/LabKey/server/pull/937

#### Changes
* Remove dev properties for configuring management endpoints

### DIFF
--- a/server/configs/application.properties
+++ b/server/configs/application.properties
@@ -114,15 +114,6 @@ mail.smtpUser=@@smtpUser@@
 # HTTP session timeout for users - defaults to 30 minutes
 #server.servlet.session.timeout=30m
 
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):
Do we also need to remove some Gradle targets?

labkey-tchad:
> Do we also need to remove some Gradle targets?

I'd like to update those tasks to shutdown the server using its PID rather than removing them altogether.

PR #1113, opened 2025-06-30
comment out audit log disabling blocks

labkey-willm:
#### Rationale
audit logging options change based on which environment we are deploying to, and these default "offs" are taking precedence over downstream overrides. Best to comment them out, but leave as an example.

#### Related Pull Requests
* https://github.com/LabKey/syseng-chef-server/pull/636

#### Changes
*

### DIFF
--- a/server/embedded/src/main/resources/log4j2.xml
+++ b/server/embedded/src/main/resources/log4j2.xml
@@ -300,7 +300,7 @@
             <AppenderRef ref="QUERY_STATS"/>
         </Logger>
 
-        <!-- <Logger name="org.labkey.audit.event" level="OFF">
+        <!-- <Logger name="org.labkey.audit.event" level="INFO">
             <AppenderRef ref="LABKEY_AUDIT"/>
         </Logger> -->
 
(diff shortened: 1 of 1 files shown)


server/embedded/src/main/resources/log4j2.xml
```diff
@@ -300,9 +300,9 @@
             <AppenderRef ref="QUERY_STATS"/>
         </Logger>
 
-        <Logger name="org.labkey.audit.event" level="OFF">
+        <!-- <Logger name="org.labkey.audit.event" level="OFF">
```
labkey-adam: Since the example is commented out, it might be more instructive to specify `INFO`.

server/embedded/src/main/resources/log4j2.xml
```diff
@@ -332,10 +332,10 @@
          individual stanza like the one shown below, where you prefix the audit event type
          with 'org.labkey.audit.event'. Then set the level to INFO and specify the appenders.
         -->
-        <Logger name="org.labkey.audit.event.UserAuditEvent" additivity="false" level="OFF">
+        <!-- <Logger name="org.labkey.audit.event.UserAuditEvent" additivity="false" level="OFF">
```
labkey-adam: `INFO`?

labkey-adam (approved):
Could specify `INFO` in commented out blocks (optional)

PR #1123, opened 2025-07-10
restore auditlog top level

labkey-willm:
#### Rationale
previous PR https://github.com/LabKey/server/pull/1113 results in audit logs going to main labkey.log by default, which is not great, and also caused certain tests to fail. This restores just the top level audit log config to send those logs to the audit log, as before.

#### Related Pull Requests
* https://github.com/LabKey/server/pull/1113

#### Changes
*


labkey-willm:
forgot to branch from snapshot

PR #1124, opened 2025-07-10
restore top level audit logging in server default log4j

labkey-willm:
#### Rationale
previous PR https://github.com/LabKey/server/pull/1113 results in audit logs going to main labkey.log by default, which is not great, and also caused certain tests to fail. This restores just the top level audit log config to send those logs to the audit log, as before.

#### Related Pull Requests
* https://github.com/LabKey/server/pull/1113

#### Changes
*

### DIFF
--- a/server/embedded/src/main/resources/log4j2.xml
+++ b/server/embedded/src/main/resources/log4j2.xml
@@ -300,7 +300,7 @@
             <AppenderRef ref="QUERY_STATS"/>
         </Logger>
 
-        <Logger name="org.labkey.audit.event" level="INFO">
+        <Logger name="org.labkey.audit.event" level="OFF">
             <AppenderRef ref="LABKEY_AUDIT"/>
         </Logger>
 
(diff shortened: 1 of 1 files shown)


server/embedded/src/main/resources/log4j2.xml
```diff
@@ -300,9 +300,9 @@
             <AppenderRef ref="QUERY_STATS"/>
         </Logger>
 
-        <!-- <Logger name="org.labkey.audit.event" level="INFO">
+        <Logger name="org.labkey.audit.event" level="INFO">
```
labkey-tchad: The previous behavior was to leave this OFF by default
```suggestion
        <Logger name="org.labkey.audit.event" level="OFF">
```

labkey-tchad (changes requested):
I don't think we want this enabled by default

labkey-tchad (approved):

PR #1125, opened 2025-07-14
Update a couple dependencies

labkey-adam:
#### Rationale
Update to the latest. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53453


labkey-susanh (commented):
Looks like there's a need to update gson too?

labkey-adam:
> Looks like there's a need to update gson too?

Yup, mentioned in https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53453 and fix committed

labkey-susanh (approved):
Looks good to me.

PR #1127, opened 2025-07-15
Gradle 8.14.3

labkey-susanh:
#### Rationale
Latest [release](https://docs.gradle.org/8.14.3/release-notes.html)

#### Changes
* Gradle version update


labkey-jeckels (approved):
FWIW, my initial build after switching to this FB took 30 minutes. Subsequent builds were in line with normal incremental builds.

labkey-susanh:
> FWIW, my initial build after switching to this FB took 30 minutes. Subsequent builds were in line with normal incremental builds.

Yes, that is expected. When there's a new version of Gradle, it can't reuse any of the caching it has done before and all tasks are considered out of date.

PR #1154, opened 2025-08-21
Suppress CVE-2024-7254

labkey-martyp:
#### Rationale
This is not a risk outside of one client specific module which is waiting on a package update using this version as a transitive dependency.

#### Changes
* Suppress cve


labkey-adam (approved):

labkey-martyp:
https://github.com/LabKey/server/pull/1155

PR #1161, opened 2025-08-30
Update tika version

labkey-martyp:
#### Rationale
CVE-2025-54988

#### Changes
* Bump to recommended version


labkey-adam:
Test failures indicate we need to update commons-compress as well: https://issues.apache.org/jira/browse/TIKA-4469. I'll address.

labkey-adam:
> Test failures indicate we need to update commons-compress as well: https://issues.apache.org/jira/browse/TIKA-4469. I'll address.

We've already got the latest commons-compress on develop. In 25.7, there are a few other dependencies with lower versions than what Tika 3.2.2 lists (https://github.com/apache/tika/blob/3.2.2/tika-parent/pom.xml), but I think we can be judicious and update only if we see problems. We may need bump more than compress on 25.3, since that's on a much earlier version of Tika.

labkey-adam:
BVTs also passed: https://teamcity.labkey.org/buildConfiguration/LabKey_257Release_Community_Bvt/3646741

labkey-adam (approved):

PR #1172, opened 2025-09-04
Issue 53723: ClassNotFoundException: JspServlet with additionalWebapp

labkey-jeckels:
#### Rationale
We don't need JSPs for additional web apps.

#### Changes
* Manually register the default servlet and other config, but don't register the JSP servlet

- [x] Manual Testing @labkey-adam 
  - Use application.properties to configure an additional webapp
  - Create a simple JSP file in the additional webapp directory (a file with just `<%= 5 + 3 %>` works)
  - Ensure file is served up as simple text instead of rendering


labkey-jeckels:
I recommend reviewing org.apache.catalina.startup.Tomcat.initWebappDefaults() to ensure that I copied the right initialization code - not too much, not too little.

labkey-adam (approved):
Reviewed against Tomcat sources and manually tested

PR #1183, opened 2025-09-17
 suppress CVE-2025-41249

labkey-willm:
#### Rationale
 We don't use @EnableMethodSecurity, so we are not impacted by CVE-2025-41249

#### Related Pull Requests
* 

#### Changes
*


labkey-jeckels:
Even when we're not affected, we typically upgrade to the patched version of the library in our current ESR (25.7 at the moment) instead of suppressing it. This means we don't have to prove that we're not vulnerable when a client's scan sees that we reference a library with a CVE.

PR #1217, opened 2025-11-17
Gradle 9.2.1 and update `vcs.xml` template to point to new location for issues

labkey-susanh:
#### Rationale
[Gradle 9.2.1 ](https://docs.gradle.org/9.2.1/release-notes.html) 

Our migration from the LabKey issue tracker into Github issue tracker put issues in the `internal-issues` project instead of the `kanban` project, so we need to update our linking syntax. Once we have resolved what happens with issues that are currently represented in both places, we'll update current code references to GitHub issues accordingly.

#### Related Pull Requests
* 

#### Changes
* Update gradle version
* Update our issue navigation link to use the new location


cnathe (approved):

labkey-adam (approved):

labkey-tchad (commented):
We should remove the scrumwise pattern from vcs.xml.

labkey-tchad:
> We should remove the scrumwise pattern from vcs.xml.

Nevermind, I was looking at the wrong release.

PR #1219, opened 2025-11-19
Update IDE issue navigation links

labkey-tchad:
#### Rationale
25.11 is going to be relatively active for a couple of months. We should have issue links point to the correct repository. A couple of the existing link definitions are no longer (or never were) valid; they should be removed.

#### Related Pull Requests
* #1217 

#### Changes
* Update our issue navigation link to use the new location
* Remove invalid issue navigation links

### DIFF
--- a/.idea/configTemplates/vcs.xml
+++ b/.idea/configTemplates/vcs.xml
@@ -18,6 +18,14 @@
           <option name="linkRegexp" value="https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=$3" />
         </IssueNavigationLink>
         <IssueNavigationLink>
+          <option name="issueRegexp" value="(Epic|Story|Item)( +#?)([0-9]+)" />
+          <option name="linkRegexp" value="https://www.scrumwise.com/scrum/#/backlog-item/$3" />
+        </IssueNavigationLink>
+        <IssueNavigationLink>
+          <option name="issueRegexp" value="^([0-9]+) ?:" />
+          <option name="linkRegexp" value="https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=$1" />
+        </IssueNavigationLink>
+        <IssueNavigationLink>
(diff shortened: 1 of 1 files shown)


.idea/configTemplates/vcs.xml
```diff
@@ -17,17 +17,9 @@
           <option name="issueRegexp" value="(Fix|Spec|Spec ID|Spec Issue|Secure Issue|Issue)( +#?)([0-9]+)" />
           <option name="linkRegexp" value="https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=$3" />
         </IssueNavigationLink>
-        <IssueNavigationLink>
-          <option name="issueRegexp" value="(Epic|Story|Item)( +#?)([0-9]+)" />
-          <option name="linkRegexp" value="https://www.scrumwise.com/scrum/#/backlog-item/$3" />
-        </IssueNavigationLink>
```
labkey-susanh, line 23: Even though we don't have Scrumwise anymore, I've found that having the link try to go there gives me some information to understand the "vintage" of the issue. Maybe we'll remember that these nouns are the "Scrumwise" nouns, and maybe that's just me, but I also think there's no harm in leaving these.

.idea/configTemplates/vcs.xml
```diff
@@ -17,17 +17,9 @@
...
-        <IssueNavigationLink>
-          <option name="issueRegexp" value="^([0-9]+) ?:" />
-          <option name="linkRegexp" value="https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=$1" />
-        </IssueNavigationLink>
```
labkey-susanh, line 27: Are we sure this last one isn't used?

.idea/configTemplates/vcs.xml
labkey-tchad: I searched my enlistment and couldn't find anything relevant that matched either of these regexes. Maybe I'm missing something though.

labkey-susanh (approved):

PR #1226, opened 2025-12-02
GitHub Issue 692: Stop using CBC in HTTPS ciphers

labkey-jeckels:
#### Rationale
CBC ciphers are now considered weak.

#### Changes
* Set the default cipher list and TLS protocol versions in Java code, sparing admins from needing to update application.properties


labkey-adam:
@labkey-jeckels should we update the docs page to recommend not setting a list of ciphers? Or at least mention that we're setting a good list by default?

labkey-adam (approved):

labkey-adam:
Perhaps add a release note as well?

labkey-jeckels:
> Perhaps add a release note as well?

Yes. See the email thread "Upcoming change to HTTPS cipher configuration" where @labkey-steveh added this to the docs todo list.

PR #1248, opened 2026-01-01
Update log4j2 to the latest version

labkey-adam:
#### Rationale
Suppress CVE-2025-68161


labkey-jeckels (approved):
I kicked off a new run of the OWASP tool but it still failed.

PR #1259, opened 2026-01-14
Spring AI version

labkey-matthewb:
#### Rationale
set global dependency version

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -202,6 +202,8 @@ allprojects {
     configurations.configureEach
             {Configuration config ->
                 resolutionStrategy {
+                    //  spring-ai pulls a different version of antr-ST4 than the version pulled by query (via antlr)
+                    force "org.antlr:ST4:${antlrST4Version}"
                     // we force this version because we have compilation problems with version 1.9.2 of commons-beanutils, which is the version
                     // that is brought in by Gradle's conflict resolution as a result of our chosen versions for commons-validator (1.5.0) and commons-digester (1.8.1)
                     force "commons-beanutils:commons-beanutils:${commonsBeanutilsVersion}"
--- a/build.gradle
+++ b/build.gradle
@@ -76,8 +76,10 @@ allprojects {
                 utilities
             }
     configurations.configureEach {
-        // exclude log4j, which may come in transitively, from all configurations to avoid its potential vulnerabilities
+        // exclude log4j 1.x, which may come in transitively, from all configurations to avoid its potential vulnerabilities
         exclude group: "log4j", module:"log4j"
+        // exclude to ensure we don't have Log4J to SLFJ adapters or other unwanted logging libraries added to modules
+        exclude group: "org.springframework.boot", module:"spring-boot-starter-logging"
     }
     configurations.driver.setDescription("Dependencies used for Gradle SetUpProperties task")
     configurations.utilities.setDescription("Utility binaries for use on Windows platform")
--- a/build.gradle
+++ b/build.gradle
@@ -204,6 +204,11 @@ allprojects {
                 resolutionStrategy {
                     //  spring-ai pulls a different version of antr-ST4 than the version pulled by query (via antlr)
                     force "org.antlr:ST4:${antlrST4Version}"
+                    // also it pulls in a bunch of com.google.api dependencies which conflict with DiscvrLabKeyModules:SequenceAnalysis and wnprc-modules:WNPRC_EHR
+                    force "com.google.api:api-common:${googleApiVersion}"
+                    force "com.google.auth:google-auth-library-oauth2-http:${googleAuthVersion}"
+                    force "com.google.auth:google-auth-library-credentials:${googleAuthVersion}"
+
                     // we force this version because we have compilation problems with version 1.9.2 of commons-beanutils, which is the version
                     // that is brought in by Gradle's conflict resolution as a result of our chosen versions for commons-validator (1.5.0) and commons-digester (1.8.1)
                     force "commons-beanutils:commons-beanutils:${commonsBeanutilsVersion}"
--- a/gradle.properties
+++ b/gradle.properties
@@ -288,6 +288,7 @@ snappyJavaVersion=1.1.10.8
 springBootVersion=3.5.8
 # This usually matches the Spring Framework version dictated by springBootVersion
 springVersion=6.2.15
+springAiVersion=1.1.0
 
 sqliteJdbcVersion=3.51.1.0
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -293,7 +293,6 @@ springBootVersion=4.0.1
 # This usually matches the Spring Framework version dictated by springBootVersion
 springVersion=7.0.2
 springAiVersion=1.1.2
-# springAiVersion=2.0.0-M1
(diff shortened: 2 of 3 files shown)


labkey-jeckels (approved):

labkey-jeckels:
Looks like we need to resolve some conflicting dependency versions.

 org.antlr:ST4 has 2 versions as follows:
15:11:53     4.0.8  [:server:modules:platform:query]
15:11:53     4.3.4  [:server:modules:platform:api]
15:11:53   org.antlr:antlr-runtime has 2 versions as follows:
15:11:53     3.5.2  [:server:modules:platform:query]
15:11:53     3.5.3  [:server:modules:platform:api]
15:11:53   com.google.api:api-common has 2 versions as follows:
15:11:53     2.14.1  [:server:modules:DiscvrLabKeyModules:SequenceAnalysis]
15:11:53     2.47.0  [:server:modules:platform:api, :server:modules:platform:query]
15:11:53   com.google.auth:google-auth-library-oauth2-http has 2 versions as follows:
15:11:53     1.33.0  [:server:modules:platform:api, :server:modules:platform:query]
15:11:53     1.19.0  [:server:modules:DiscvrLabKeyModules:SequenceAnalysis, :server:modules:wnprc-modules:WNPRC_EHR]
15:11:53   com.google.genai:google-genai has 2 versions as follows:
15:11:53     1.28.0  [:server:modules:platform:api]
15:11:53     1.15.0  [:server:modules:platform:query]
15:11:53   com.google.auth:google-auth-library-credentials has 2 versions as follows:
15:11:53

gradle.properties
```diff
@@ -288,6 +288,7 @@ snappyJavaVersion=1.1.10.8
 springBootVersion=3.5.8
 # This usually matches the Spring Framework version dictated by springBootVersion
 springVersion=6.2.15
+springAiVersion=1.1.2
```
labkey-jeckels, line 295: Do we need this at the root? Looks like it's only used in the API module. Do you expect other modules will need it before long?

Regardless, we need to resolve the handful of dependency version conflicts before this can be merged.

labkey-jeckels (changes requested):
Need to resolve dependency conflicts

labkey-matthewb:
> Need to resolve dependency conflicts

I put springAIVersion here just to put it with its spring friends.  I'm open to doing something else.

I forced force antlrST4Version=4.3.4 (in platform).  Seems to work, but I'm not sure how we maintain and review these exceptions.

labkey-jeckels:
> > Need to resolve dependency conflicts
> 
> I put springAIVersion here just to put it with its spring friends. I'm open to doing something else.
> 
> I forced force antlrST4Version=4.3.4 (in platform). Seems to work, but I'm not sure how we maintain and review these exceptions.

I believe we primarily rely on the comments in the line that forces the version. Thanks for adding one.

@labkey-susanh any preference on where to set the Spring AI version? I thought it would live in the one module that uses it, but I don't see a `server/modules/platform/api/gradle.properties` so I'm probably wrong.

labkey-jeckels:
@labkey-matthewb that cleared one of the four conflicts. We probably need forces for one or more of the others too.

labkey-susanh:
> > > Need to resolve dependency conflicts
> > 
> > 
> > I put springAIVersion here just to put it with its spring friends. I'm open to doing something else.
> > I forced force antlrST4Version=4.3.4 (in platform). Seems to work, but I'm not sure how we maintain and review these exceptions.
> 
> I believe we primarily rely on the comments in the line that forces the version. Thanks for adding one.
> 
> @labkey-susanh any preference on where to set the Spring AI version? I thought it would live in the one module that uses it, but I don't see a `server/modules/platform/api/gradle.properties` so I'm probably wrong.

Often if there's just one usage, we won't capture it in a `gradle.properties` file but inline in the `build.gradle` file. We aren't 100% consistent with these one-offs. The main thing we want is to use a property when there is more than one usage to maintain consistency. If the Spring AI version will need to be updated along with its other spring buddies, keeping them together makes sense. We also maintain notes on current versions and latest versions and the like in the "Upgrade Dependencies" document. This includes notes on which `build.gradle` file needs updating for various dependencies.

labkey-jeckels:
> > > > Need to resolve dependency conflicts
> > > 
> > > 
> > > I put springAIVersion here just to put it with its spring friends. I'm open to doing something else.
> > > I forced force antlrST4Version=4.3.4 (in platform). Seems to work, but I'm not sure how we maintain and review these exceptions.
> > 
> > 
> > I believe we primarily rely on the comments in the line that forces the version. Thanks for adding one.
> > @labkey-susanh any preference on where to set the Spring AI version? I thought it would live in the one module that uses it, but I don't see a `server/modules/platform/api/gradle.properties` so I'm probably wrong.
> 
> Often if there's just one usage, we won't capture it in a `gradle.properties` file but inline in the `build.gradle` file. We aren't 100% consistent with these one-offs. The main thing we want is to use a property when there is more than one usage to maintain consistency. If the Spring AI version will need to be updated along with its other spring buddies, keeping them together makes sense. We also maintain notes on current versions and latest versions and the like in the "Upgrade Dependencies" document. This includes notes on which `build.gradle` file needs updating for various dependencies.

@labkey-matthewb here's the link to that "Upgrade Dependencies" doc: https://docs.google.com/document/d/1JPZ8al_M6eQh813Q8y3ALhRNgTieveaKDXv5iFUbq7A/edit?tab=t.0

build.gradle
```diff
@@ -76,8 +76,10 @@ allprojects {
                 utilities
             }
     configurations.configureEach {
-        // exclude log4j, which may come in transitively, from all configurations to avoid its potential vulnerabilities
+        // exclude log4j 1.x, which may come in transitively, from all configurations to avoid its potential vulnerabilities
         exclude group: "log4j", module:"log4j"
+        // exclude to ensure we don't have Log4J to SLFJ adapters or other unwanted logging libraries added to modules
+        exclude group: "org.springframework.boot", module:"spring-boot-starter-logging"
```
labkey-jeckels, line 82: @labkey-susanh this is how I suppressed this transitive dependency from being pulled into modules. Happy to shift to a better place if you have a recommendation.

labkey-susanh: Looks good to me.

gradle.properties
```diff
@@ -288,6 +292,8 @@ snappyJavaVersion=1.1.10.8
 springBootVersion=4.0.1
 # This usually matches the Spring Framework version dictated by springBootVersion
 springVersion=7.0.2
+springAiVersion=1.1.2
+# springAiVersion=2.0.0-M1
```
labkey-susanh: Can we get rid of this commented out version?

labkey-susanh (commented):
Looks like there's a need for a merge from develop? DRTs are failing.

labkey-jeckels:
> Looks like there's a need for a merge from develop? DRTs are failing.

Merged.

gradle.properties
labkey-jeckels: Removed.

labkey-susanh (approved):

labkey-jeckels (approved):

PR #1261, opened 2026-01-19
Gradle 9.3.0

labkey-susanh:
#### Rationale
[Gradle 9.3.0](https://docs.gradle.org/9.3.0/release-notes.html) is the latest

#### Changes
* Gradle version update


labkey-adam (approved):

labkey-tchad (commented):
I was able to prevent the `MarkedVerificationException` when running the test task but it just makes the task succeed without running any tests.

We might need to make some updates to our test runner (`org.labkey.test.Runner`). It was designed log ago to run with Ant; it seems very possible that we were depending on some corner case that is no longer supported. Gradle 9.3 did make some changes to their test reporting: https://docs.gradle.org/current/release-notes.html#test-reporting-improvements

labkey-tchad (approved):
Test update has been merged

PR #1265, opened 2026-01-23
GitHub Issue 796: JSON logging stopped after Tomcat/Spring update

labkey-jeckels:
#### Rationale
We configure JSON-based Log4J logging via an override file in our cloud instances. This stopped working as of the upgrade to Spring Boot 4.

#### Changes
* Propagate from the Log4J-preferred config to the Spring Boot-preferred config so they stop squabbling over the setup

### DIFF
--- a/server/embedded/src/org/labkey/embedded/LabKeyServer.java
+++ b/server/embedded/src/org/labkey/embedded/LabKeyServer.java
@@ -92,6 +92,32 @@ public class LabKeyServer
 
         application.setDefaultProperties(new HashMap<>()
              {{
+                // GitHub Issue 796: JSON logging stopped after Tomcat/Spring update
+                 // Propagate log4j configuration to Spring Boot config, which is necessary with Spring Boot 4.x
+                 String log4JConfig = System.getProperty("log4j.configurationFile");
+                 if (log4JConfig != null)
+                 {
+                     String[] log4JConfigParts = log4JConfig.split(",");
+                     if (log4JConfigParts.length > 0)
+                     {
+                         if ("log4j2.xml".equals(log4JConfigParts[0]))
+                         {
+                             // Assume this is the one packaged with our embedded build and on the classpath
+                             put("logging.config", "classpath:log4j2.xml");
+                         }
+                         else
+                         {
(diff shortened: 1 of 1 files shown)


labkey-willm:
Nice hack. Does this still work if there's 3 or more files in play? labkey.org and HSE (audit logs) each get a 3rd file: https://github.com/LabKey/syseng-chef-server/blob/develop/cookbooks/lk_appserver/recipes/default_embedded.rb#L785

labkey-jeckels:
> Nice hack. Does this still work if there's 3 or more files in play? labkey.org and HSE (audit logs) each get a 3rd file: https://github.com/LabKey/syseng-chef-server/blob/develop/cookbooks/lk_appserver/recipes/default_embedded.rb#L785

It didn't originally but does now. I tested with three, and it should be fine with whatever number now.

We should transition from using the Log4J-style config to the Spring Boot config, but this helps us make the transition.

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -92,6 +92,40 @@ public static void main(String[] args)
...
+                         {
+                             StringBuilder sb = new StringBuilder();
+                             String separator = "";
+                             for (int i = 1; i < log4JConfigParts.length; i++)
+                             {
+                                 sb.append(separator);
+                                 separator = ",";
+                                 sb.append(log4JConfigParts[i]);
+                             }
+                            put("logging.log4j2.config.override", sb.toString());
```
labkey-adam: You could replace the `StringBuilder` and loop with a one-liner:

```suggestion
                            put("logging.log4j2.config.override", String.join(",", log4JConfigParts));
```

labkey-adam (approved):

server/embedded/src/org/labkey/embedded/LabKeyServer.java
labkey-jeckels: Replaced, but with a longer one-liner. This needs to exclude the first path in the list.

labkey-adam: Right, missed that.

PR #1284, opened 2026-02-16
Improved setup for XML parsing

labkey-jeckels:
#### Rationale
We should configure XML parsers consistently. Code in this repo can't leverage XmlBeanUtil so do it manually.

#### Changes
* Apply standard settings to the SAX parser

### DIFF
--- a/server/bootstrap/src/org/labkey/bootstrap/ModuleArchive.java
+++ b/server/bootstrap/src/org/labkey/bootstrap/ModuleArchive.java
@@ -72,7 +72,7 @@ public class ModuleArchive
 
         try
         {
-            // Keep this in sync with config on XmlBeanUtil.SAX_PARSER_FACTORY. See motiviations in comments there.
+            // Keep this in sync with config on XmlBeansUtil.SAX_PARSER_FACTORY. See motiviations in comments there.
             SAXParserFactory factory = SAXParserFactory.newDefaultInstance();
             factory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
             factory.setFeature("http://xml.org/sax/features/external-general-entities", false);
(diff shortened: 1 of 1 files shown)


server/bootstrap/src/org/labkey/bootstrap/ModuleArchive.java
```diff
@@ -71,7 +71,12 @@ private String nameFromModuleXML(InputStream is) throws IOException
 
         try
         {
-            SAXParser parser = SAXParserFactory.newDefaultInstance().newSAXParser();
+            SAXParserFactory factory = SAXParserFactory.newDefaultInstance();
+            factory.setFeature(javax.xml.XMLConstants.FEATURE_SECURE_PROCESSING, true);
+            factory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
```
labkey-alan, line 77: According to this blog post I found from 2012 these settings seem to be mutually exclusive. My understanding is that you need to either disable doctype declaration (what this line does) *or* disable external entities (what the following two lines do). I'm not sure it's problematic to do both though, so maybe we should keep it.

https://blog.compass-security.com/2012/08/secure-xml-parser-configuration/

labkey-alan (approved):
Overall this looks good to me, however, I did have to search the web to find an explanation for why we want to do this. I think it would be a good idea to add a comment explaining what we're doing and why. I also left a more specific comment, but I'm not sure it's important enough to prevent merging.

server/bootstrap/src/org/labkey/bootstrap/ModuleArchive.java
labkey-jeckels: Yes, should be safe to do both, just in case.

labkey-jeckels:
> Overall this looks good to me, however, I did have to search the web to find an explanation for why we want to do this. I think it would be a good idea to add a comment explaining what we're doing and why. I also left a more specific comment, but I'm not sure it's important enough to prevent merging.

Added two-way pointers here and in XmlBeansUtil, improved comments, and made them more consistent.

PR #1291, opened 2026-02-20
Ensure consistent XML parser config

labkey-jeckels:
#### Rationale
We want to be using our centrally defined XML parser configuration

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7440

#### Changes
- Inspection to discourage non-standard XML parser factory configs
- Suppress some less useful inspections

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -1,9 +1,8 @@
 <component name="InspectionProjectProfileManager">
-  <profile version="1.0" is_locked="false">
+  <profile version="1.0">
     <option name="myName" value="Project Default" />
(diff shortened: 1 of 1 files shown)


labkey-adam (commented):
Unfortunate that IntelliJ (I'm assuming) rearranged everything in the inspection file. On quick glance, it looks like a couple inspections got removed: e.g., AutoCloseableResource and WeakerAccess. Intentional? Maybe they're on by default so don't need to be specified?

labkey-jeckels:
> Unfortunate that IntelliJ (I'm assuming) rearranged everything in the inspection file. On quick glance, it looks like a couple inspections got removed: e.g., AutoCloseableResource and WeakerAccess. Intentional? Maybe they're on by default so don't need to be specified?

AutoCloseable and WeakerAccess appear to be redundant with current IntelliJ defaults.

The Unicode change was unintentional and has been reverted.

The JavaDoc ones were intentional.

The custom inspections appear to be have been alphabetized by name, by IntelliJ. Hopefully that means they'll be easier to diff going forward.

labkey-adam (approved):

PR #1292, opened 2026-02-23
AI front end code review guidelines and claude skills

labkey-martyp:
#### Rationale
This is a starting point for frontend ai code review guidelines. There are two code review checklists and claude skills, /code-review-react and /code-review-jest. The claude skills have two modes, 
1. Review only staged changes (changes not committed yet). ex. `/code-review-react <directory or file path>`
2. Review full files or directories. ex `/code-review-jest --full <directory or file path>` 

- See .agents/common.md for review priority and format.
- See the "Required output" section of the `.claude/skills/<skill name>/skill.md` for the expected output. Note: If there are more than 10 "suggested" changes (non-urgent), only the first 10 will be shown in each run.
- See the files in `.agents/review-checklists/<react or jest>/*.md` for the code review checks.  

It is recommended to run the code reviews multiple times until all the feedback is addressed.

These guidelines can be used by other ai agents as well, just point them to the directory.

#### Changes
- .agents/review-checklists/jest code review checks
- .agents/review-checklists/react code review checks
- .agents/review-checkslists/common.md general code review guidelines
- `.claude/skill/<skill name>/skill.md` define the claude skill using these guidelines

### DIFF
--- a/.agents/review-checklists/jest/business-logic.md
+++ b/.agents/review-checklists/jest/business-logic.md
@@ -17,7 +17,6 @@ Flag as a high-severity finding when the test would still pass after removing th
 ### Exceptions / False Positives
 
 - Allow explicit smoke tests when the purpose is only to verify the component/function does not throw on render or initialization.
-- Do not flag helper tests that intentionally validate test utilities/builders rather than product behavior, if the subject under test is the utility itself.
 - A simple existence assertion can be acceptable when the behavior under review is conditional presence/absence itself (for example, permission-gated rendering).
 
 ### Detection heuristic
@@ -53,6 +52,7 @@ expect(data[0].name).toBe('Alpha'); // This tests your test, not your code
 // ✅ GOOD — asserts on rendered output from that input
 const data = [{ name: 'Alpha', value: 10 }];
 render(<Table rows={data} />);
-expect(screen.getByText('Alpha')).toBeInTheDocument();
-expect(screen.getByText('10')).toBeInTheDocument();
+const row = screen.getByRole('row', { name: /alpha\s+10/i });
+expect(within(row).getByRole('cell', { name: 'Alpha' })).toBeInTheDocument();
+expect(within(row).getByRole('cell', { name: '10' })).toBeInTheDocument();
 ```
--- a/.agents/review-checklists/jest/business-logic.md
+++ b/.agents/review-checklists/jest/business-logic.md
@@ -1,0 +1,58 @@
+# Business Logic
+
+> **Prerequisite:** Review and apply the common guidelines in [`common.md`](../common.md) before using this checklist.
+
+## Jest tests must assert on meaningful outcomes
+
+**Urgency:** urgent
+
+### Category
+
+Correctness
+
+### Confidence Threshold
+
+Flag as a high-severity finding when the test would still pass after removing the feature logic, or when assertions only validate setup/static existence. If the test is intentionally a smoke/no-crash render test, require that intent to be explicit in the test name.
+
+### Exceptions / False Positives
+
+- Allow explicit smoke tests when the purpose is only to verify the component/function does not throw on render or initialization.
+- Do not flag helper tests that intentionally validate test utilities/builders rather than product behavior, if the subject under test is the utility itself.
+- A simple existence assertion can be acceptable when the behavior under review is conditional presence/absence itself (for example, permission-gated rendering).
+
+### Detection heuristic
+
+Look for tests where all assertions target mock setup data, static strings, or DOM existence rather than computed/rendered output.
+
+## Rules
+
+1. **Assert on component output, not existence.** After `render()`, assert on visible text, element states, or DOM changes that result from the props/state you set up — never just `expect(container).toBeDefined()` or `expect(document.querySelector('.x')).not.toBeNull()`.
+
+2. **Assert on computed results, not test inputs.** If you pass `mockData` into a component, don't assert that `mockData` has the values you just wrote. Assert on what the component *did* with that data.
+
+3. **Every test must fail if the feature is removed.** Apply this litmus test: if you deleted the implementation code for the feature under test, would the test still pass? If yes, the test is worthless — rewrite it.
+
+4. **Interact before asserting (when applicable).** If testing behavior triggered by user action (click, submit, input), simulate that action with `fireEvent` or `userEvent`, then assert on the resulting DOM or state change.
+
+## Examples
+
+```js
+// ❌ BAD — asserts on render existence
+render(<MyForm valid={false} />);
+expect(document.querySelector('.form-container')).not.toBeNull();
+
+// ✅ GOOD — asserts on behavioral outcome of props
+render(<MyForm valid={false} />);
+expect(screen.getByRole('button', { name: /submit/i })).toBeDisabled();
+
+// ❌ BAD — asserts on mock input
+const data = [{ name: 'Alpha', value: 10 }];
+render(<Table rows={data} />);
+expect(data[0].name).toBe('Alpha'); // This tests your test, not your code
+
+// ✅ GOOD — asserts on rendered output from that input
+const data = [{ name: 'Alpha', value: 10 }];
+render(<Table rows={data} />);
+expect(screen.getByText('Alpha')).toBeInTheDocument();
+expect(screen.getByText('10')).toBeInTheDocument();
+```
--- a/.agents/review-checklists/jest/code-quality.md
+++ b/.agents/review-checklists/jest/code-quality.md
@@ -1,0 +1,208 @@
+# Code Quality
+
+> **Prerequisite:** Review and apply the common guidelines in [`common.md`](../common.md) before using this checklist.
+
+## Arrange / Act / Assert (AAA) structure
+
+**Urgency:** suggestion
+
+### Category
+
+Style
+
+### Confidence Threshold
+
+Flag only when the test is long enough or complex enough that structure affects readability (for example, multi-step setup/interactions or >10 lines). For short, self-evident tests, prefer a suggestion or no comment.
+
+### Exceptions / False Positives
+
+- Do not flag short tests (roughly 3-5 lines) where Arrange/Act/Assert is obvious without comments.
+- Do not require AAA comments in parameterized tests when added comments make the test harder to scan than the code itself.
+- Prefer suggestions over high-severity findings unless the repository explicitly enforces AAA comment structure in tests.
+
+### Rules
+
+1. **Each section must be preceded by a comment** — `// Arrange`, `// Act`, and `// Assert`.
+2. **The Assert comment must include a brief explanation** of how the described scenario is being verified. The explanation should summarize what the assertions prove about the behavior stated in the test name.
+   - Good: `// Assert - textarea shows hash subjects, participantId query param is ignored`
+   - Good: `// Assert - ID Search button is active`
+   - Bad: `// Assert` (missing explanation)
+   - Bad: `// Assert - check results` (too vague; does not connect to the scenario)
+3. **Arrange may be omitted** when there is no setup beyond what `beforeEach` already provides, but Act and Assert are always required.
+4. **Combined `// Act & Assert` is acceptable** only when the assertion must be wrapped inside a `waitFor` that is inseparable from the action (e.g., awaiting an async side-effect). In that case the comment must still include an explanation:
+   - Good: `// Act & Assert - empty state message is shown and error was logged to console`
+   - Bad: `// Act & Assert`
+5. **Multiple Act/Assert cycles in one test are allowed** for stateful interaction flows (e.g., click then verify, click again then verify). Each cycle must carry its own `// Act` and `// Assert - ...` comments.
+
+---
+
+## No unused variables, imports, or mocks
+
+**Urgency:** urgent
+
+### Category
+
+Maintainability
+
+### Confidence Threshold
+
+Flag when the symbol is clearly unused in the file or when a mock setup is not consumed by behavior under test. If a mock or import may be used implicitly (module mocking side effects, global setup conventions), verify before commenting.
+
+### Exceptions / False Positives
+
+- Do not flag side-effect imports (for example, polyfills or test environment setup) just because no identifier is referenced.
+- Do not flag module-level `jest.mock(...)` declarations that intentionally replace imports used elsewhere in the file.
+- If a placeholder parameter is required for function signature/position, allow underscore-prefixed names (for example, `_arg`).
+
+### Rules
+
+1. **Unused imports must be removed.** If a symbol is imported but never referenced in the file, delete the import. This includes named imports, default imports, and type-only imports.
+2. **Unused variables and constants must be removed.** If a `const`, `let`, or destructured binding is declared but never read, delete it. This applies to top-level declarations, inside `describe`/`beforeEach`/`test` blocks, and helper functions.
+3. **Unused mock declarations must be removed.** If `jest.fn()`, `jest.mock()`, `jest.spyOn()`, or a manual mock variable is set up but never referenced in an assertion or as a dependency, delete it. Mocks that are called implicitly (e.g., module-level `jest.mock('...')` that replaces an import used elsewhere) are considered used.
+4. **Unused mock return values must be removed.** If a mock is configured with `.mockReturnValue()`, `.mockResolvedValue()`, or `.mockImplementation()` but the return value is never consumed or asserted on, simplify or remove the configuration.
+5. **Unused helper functions and factory functions must be removed.** If a test utility, builder, or factory function defined in the file is never called, delete it.
+6. **Unused parameters in callbacks must be prefixed with `_`.** If a callback parameter (e.g., in `.mockImplementation((unusedArg) => ...)`) is required for positional reasons but not used, prefix it with `_` to signal intent (e.g., `_unusedArg`).
+7. **Unused `render` results must not be destructured.** If `render(<Component />)` is called and the return value is not used, do not destructure it. Write `render(<Component />);` instead of `const { container } = render(<Component />);` when `container` is never referenced.
+
+### Examples
+
+```tsx
+// ---- Unused imports ----
+
+// ❌ BAD — ApiResponse is imported but never used
+import { render, screen } from '@testing-library/react';
+import { ApiResponse, UserProfile } from '../models';
+
+const mockProfile: UserProfile = { name: 'Test' };
+
+// ✅ GOOD — only referenced imports remain
+import { render, screen } from '@testing-library/react';
+import { UserProfile } from '../models';
+
+const mockProfile: UserProfile = { name: 'Test' };
+
+
+// ---- Unused variables ----
+
+// ❌ BAD — mockHandler is declared but never used
+const mockHandler = jest.fn();
+const mockCallback = jest.fn();
+
+test('calls callback on click', () => {
+    render(<Button onClick={mockCallback} />);
+    fireEvent.click(screen.getByRole('button'));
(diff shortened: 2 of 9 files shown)


labkey-jeckels (approved):
I don't have React or Jest changes handy to test this out, but the overall structure seems reasonable.

Others can better comment on whether these are the right items to highlight, but this might be best evaluated through actual testing.

I can see pulling some of this guidance into more general areas in the future. For example, the rules for code comments should be applicable to any language, not just React/TypeScript/JavaScript.

.claude/skills/code-review-jest/skill.md
```diff
@@ -0,0 +1,123 @@
+---
+name: code-review-jest
+description: "Trigger when the user requests a review of Jest test files (e.g., `.test.tsx`, `.spec.tsx`). Supports --full flag for complete file/directory review; defaults to staged changes when a path is provided."
```
cnathe: we can remove the `.spec.tsx` part right? those are dead tests that we need to convert. Instead maybe make the files target `.test.tsx` and `.test.ts`?

.claude/skills/code-review-jest/skill.md
```diff
@@ -0,0 +1,123 @@
...
+---
+
+# Jest Test Code Review
+
+## Intent
+Use this skill whenever the user asks to review Jest test code (especially `.test.tsx`, `.spec.tsx`, `.test.ts`, or `.spec.ts` files). Support three review modes:
```
cnathe: same comment here about spec files

.claude/skills/code-review-react/skill.md
```diff
@@ -0,0 +1,123 @@
...
+
+**Multi-file grouping:** When reviewing multiple files, group all findings together by urgency section (urgent issues first, then suggestions), not per-file. Include the file path in each finding's `FilePath` field.
+
+1. Open the relevant component/module. Gather all lines.
+2. For each applicable checklist rule, evaluate the code against the rule text, confidence threshold, and exceptions/false positives before deciding to flag it.
+3. For each confirmed violation, capture evidence (exact snippet and/or file/line), record the rule's primary category, and note confidence briefly.
+4. Compose the review section per the template below. Group findings by **Urgency** section first (urgent issues, then suggestions). Within each section, order findings by the checklist primary category priority: **Correctness**, then **Maintainability**, then **Style**.
+
+## Required output
+When invoked, the response must exactly follow one of the two templates:
```
cnathe, line 72: so much of the text / instructions above this line in this file are so close to a duplicate of the code-review-react/skill.md file. This likely makes sense as there might not be a good way to share that part of these instructions. I'm just a little worried about how we are goign to keep these things up to date and in sync.

.agents/review-checklists/jest/code-quality.md
```diff
@@ -0,0 +1,208 @@
...
+
+**Urgency:** urgent
+
+### Category
+
+Maintainability
+
+### Confidence Threshold
+
+Flag when the symbol is clearly unused in the file or when a mock setup is not consumed by behavior under test. If a mock or import may be used implicitly (module mocking side effects, global setup conventions), verify before commenting.
```
cnathe, line 49: what does "the symbol" mean in this sentence? do you mean variable?

.agents/review-checklists/jest/code-quality.md
```diff
@@ -0,0 +1,208 @@
...
+4. **Unused mock return values must be removed.** If a mock is configured with `.mockReturnValue()`, `.mockResolvedValue()`, or `.mockImplementation()` but the return value is never consumed or asserted on, simplify or remove the configuration.
+5. **Unused helper functions and factory functions must be removed.** If a test utility, builder, or factory function defined in the file is never called, delete it.
+6. **Unused parameters in callbacks must be prefixed with `_`.** If a callback parameter (e.g., in `.mockImplementation((unusedArg) => ...)`) is required for positional reasons but not used, prefix it with `_` to signal intent (e.g., `_unusedArg`).
+7. **Unused `render` results must not be destructured.** If `render(<Component />)` is called and the return value is not used, do not destructure it. Write `render(<Component />);` instead of `const { container } = render(<Component />);` when `container` is never referenced.
+
+### Examples
```
cnathe, line 67: this is great. I'd love to not have to manually check as a code reviewer if the import or variable is used!

.agents/review-checklists/jest/performance.md
```diff
@@ -0,0 +1,105 @@
+# Performance
+
+> **Prerequisite:** Review and apply the common guidelines in [`common.md`](../common.md) before using this checklist.
+
+## No redundant or near-duplicate tests
```
cnathe, line 5: I'm curious what AI would consider a near-duplicate test. I think using near duplicate tests is exactly how test cases often look when you are testing interactions of a sequence of parameters or properties (i.e. adjust prop1, then prop2, then prop3, etc. but the test is nearly the same).

.agents/review-checklists/react/business-logic.md
```diff
@@ -0,0 +1,248 @@
...
+
+Always initialize state with a value matching the controlled type:
+
+```tsx
+// ✅ Explicit empty string for text inputs
+const [input, setInput] = useState('');
+
+// ✅ Fallback for optional props
+const [text, setText] = useState(props.initialText ?? '');
+```
```
cnathe, line 243: this is great. these are always tough things to catch when manually scrolling through a large PR code change.

.agents/review-checklists/react/code-quality.md
```diff
@@ -0,0 +1,462 @@
...
+    report: ReportConfig;
+    tab: ExtReportTab;
+}
+
+const MyComponent: FC<MyComponentProps> = ({ report, tab }) => {
+    // ...
+};
+```
+
+## Optional props that are always passed should be required
```
cnathe, line 46: this one has been debated a few times before by the Frontend group and I think people fall on different camps here. Seems fine to include as long as we are making an effort to be sure to update / adjust this set of rules as we work through usages.

.agents/review-checklists/react/code-quality.md
```diff
@@ -0,0 +1,462 @@
...
+3. Flag class names that start with `--` or `__` without a preceding block name — modifiers and elements must always include their block.
+4. Look for generic class names like `container`, `wrapper`, `item`, `box` without a block prefix — these should be scoped to a BEM block.
+
+### Exceptions
+
+- **Third-party / framework classes:** Bootstrap (`btn`, `btn-primary`, `col-xs-*`, `panel-body`, `form-control`, etc.), LabKey API classes (`labkey-*`), or other library-provided class names follow their own conventions and must not be flagged or renamed.
+- **CSS Modules:** When using CSS Modules (`.module.css` / `.module.scss`), camelCase class names are standard since they become JS property names (e.g., `styles.searchForm`, `styles.activeItem`). Do not flag these as BEM violations.
+- **Existing non-BEM code:** Class names in unchanged code that predate BEM adoption should not be flagged. Only apply BEM conventions to new or actively modified class names in the PR.
+
+## No orphaned or unused CSS/SCSS rules
```
cnathe, line 402: this would be great if AI can help with this one! this is very tedious to track down manually

cnathe (approved):
I left some comments but I agree that this will need to be treated as a working document as we use it and until we start using it, it is hard to know exactly what we all want here. Looks like a great starting point.

.claude/skills/code-review-jest/skill.md
labkey-alan: I agree, if anything we should probably have another skill for converting enzyme tests to RTL, but I also don't think it's high priority enough require it go in this PR.

.agents/review-checklists/jest/performance.md
labkey-alan: So, in those cases we should be using something like `jest.each`, where we have one shared test body, but you parametrize the test with an array of args. We don't do this a lot right now, but we should for sure be doing it more.

.agents/review-checklists/react/code-quality.md
labkey-alan: I don't think this has been debated. I'm the one who has brought this to the group multiple times, basically anytime I've encountered an issue caused by this, or when I've done large refactors and discovered multiple type related issues, and I don't recall a time that I have gotten push back on this. The upside is that by following this rule your types are more correct, and it eliminates a whole class of bug, the downside is that it's slightly more verbose.

.agents/review-checklists/jest/business-logic.md
```diff
@@ -0,0 +1,58 @@
...
+Correctness
+
+### Confidence Threshold
+
+Flag as a high-severity finding when the test would still pass after removing the feature logic, or when assertions only validate setup/static existence. If the test is intentionally a smoke/no-crash render test, require that intent to be explicit in the test name.
+
+### Exceptions / False Positives
+
+- Allow explicit smoke tests when the purpose is only to verify the component/function does not throw on render or initialization.
+- Do not flag helper tests that intentionally validate test utilities/builders rather than product behavior, if the subject under test is the utility itself.
```
labkey-alan: I am not sure I understand this rule. Are we ever testing our test utilities?

.agents/review-checklists/jest/business-logic.md
```diff
@@ -0,0 +1,58 @@
...
+render(<MyForm valid={false} />);
+expect(screen.getByRole('button', { name: /submit/i })).toBeDisabled();
+
+// ❌ BAD — asserts on mock input
+const data = [{ name: 'Alpha', value: 10 }];
+render(<Table rows={data} />);
+expect(data[0].name).toBe('Alpha'); // This tests your test, not your code
+
+// ✅ GOOD — asserts on rendered output from that input
+const data = [{ name: 'Alpha', value: 10 }];
```
labkey-alan, line 53: I wonder if we should encourage even more specific checks here. I think keeping this section is good, but maybe adding something that says "instead of looking for the existence of data anywhere, look for the existence of data where it is expected to be". Then provide an example of looking for "name" to be in the first column of the example Table component here, or something similar. Just generically looking for content to be on the page may be so overbroad that it is not useful.

.agents/review-checklists/jest/code-quality.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Confidence Threshold
+
+Flag only when the test is long enough or complex enough that structure affects readability (for example, multi-step setup/interactions or >10 lines). For short, self-evident tests, prefer a suggestion or no comment.
+
+### Exceptions / False Positives
+
+- Do not flag short tests (roughly 3-5 lines) where Arrange/Act/Assert is obvious without comments.
+- Do not require AAA comments in parameterized tests when added comments make the test harder to scan than the code itself.
+- Prefer suggestions over high-severity findings unless the repository explicitly enforces AAA comment structure in tests.
```
labkey-alan, line 21: I am not sure this is useful, since we don't have any repositories where AAA comments are explicitly enforced, and most of our existing tests do not follow this pattern.

.agents/review-checklists/react/business-logic.md
```diff
@@ -0,0 +1,248 @@
...
+- Stale data displayed after list mutations
+
+### Suggested Fix
+
+Use a stable, unique identifier from the data itself:
+
+- **Before:** `items.map((item, index) => <Row key={index} ... />)`
+- **After:** `items.map(item => <Row key={item.id} ... />)`
+
+If no natural unique key exists, consider whether the data model should include one. As a last resort, generate stable IDs when the data is first loaded.
```
labkey-alan: This line feels redundant given the (IMO) more thorough instructions in the `Exceptions / False Positives` section above.

.agents/review-checklists/react/business-logic.md
```diff
@@ -0,0 +1,248 @@
...
+### Suggested Fix
+
+Convert the condition to a boolean explicitly:
+
+```tsx
+// ✅ Renders nothing when count is 0
+{count > 0 && <Badge count={count} />}
+
+// ✅ Or use a ternary
+{count ? <Badge count={count} /> : null}
```
labkey-alan: This pattern is always worse IMO and I don't think we should offer it as an acceptable alternative. The `count > 0 &&` solution above is less verbose and more explicit, which is better.

.agents/review-checklists/react/code-quality.md
```diff
@@ -0,0 +1,462 @@
...
+Style
+
+### Confidence Threshold
+
+Flag only when the component has multiple props or the inline type is large enough to reduce readability. For tiny local components with one simple prop, prefer no comment unless the file is being standardized.
+
+### Exceptions / False Positives
+
+- Do not flag tiny local components with a single simple prop where an inline type is shorter and clearer.
+- Do not require this pattern if the surrounding file consistently uses a different accepted convention and the PR is not refactoring style.
```
labkey-alan: I think this exception should be removed. I don't believe this is true anywhere in our code, and even if it was, I would want folks to update old code that doesn't meet this pattern if they're touching the file.

.agents/review-checklists/react/code-quality.md
```diff
@@ -0,0 +1,462 @@
...
+- If the project avoids `FC` entirely and types props on the function parameter instead, apply the spirit of the rule (named props type) rather than the exact syntax.
```
labkey-alan: I think we should remove this rule. We require components to be typed, so if someone made a component that was not labeled with `FC` I would consider that bad, and would request that is fixed prior to merging.

.agents/review-checklists/react/code-quality.md
```diff
@@ -0,0 +1,462 @@
...
+}
+```
+
+### How to Check
+
+1. Search for all usages of the component across the codebase
+2. For each optional prop, verify whether any call site omits it
+3. If all call sites provide the prop, flag it for conversion to required
+
+**Scope note:** For widely-used shared components (5+ call sites), this check can be deferred to a separate audit. Focus on components with 1–3 usages where the call sites are visible in the current PR.
```
labkey-alan:
```suggestion
**Scope note:** For widely-used shared components (5+ call sites), this check can be deferred to a separate audit. Focus on components with 1–4 usages where the call sites are visible in the current PR.
```

.agents/review-checklists/react/code-quality.md
```diff
@@ -0,0 +1,462 @@
...
+
+**Urgency:** suggestion
+
+### Category
+
+Style
+
+### Confidence Threshold
+
+Flag only when the file/area already uses a class-name utility convention (for example, `classnames`, `clsx`, or local `cn`) and the inline conditional harms consistency/readability. For one-off simple cases in mixed-style files, prefer a suggestion.
```
labkey-alan, line 263: We only use `classnames`, and unless someone has a good reason to add a new dependency, I don't see that changing

.agents/review-checklists/react/performance.md
```diff
@@ -0,0 +1,144 @@
...
+
+**Urgency:** suggestion
+
+### Category
+
+Maintainability
+
+### Confidence Threshold
+
+Flag when the computation is non-trivial (for example, sorting/filtering large collections or repeated transformations) and can run frequently with unchanged inputs. For small arrays or infrequent renders, make this a suggestion only.
```
labkey-alan:
```suggestion
Flag when the computation is non-trivial (for example, sorting/filtering large collections or repeated transformations) and can run frequently with unchanged inputs. For small arrays or infrequent renders, make this a suggestion only. If array size is not known, prefer memoization.
```

.agents/review-checklists/react/performance.md
```diff
@@ -0,0 +1,144 @@
...
+
+**Urgency:** urgent
+
+### Category
+
+Correctness
+
+### Confidence Threshold
+
+Flag as a high-severity finding when a nested function is used as a React component type (capitalized and rendered with JSX) inside a component body, causing remounts and state loss. Do not confuse this with render helpers that return JSX but are not treated as component types.
```
labkey-alan, line 110: I think we should consider a rule that discourages render helpers. We have relied on them a ton in the past, but not in helpful ways. 99/100 render helpers at LabKey should be components.

labkey-alan (approved):
I think overall these rules are fine. I did leave some feedback that you should review before merging. If you have follow up questions, or want to have a discussion, let me know.

My biggest issue with these rules at the moment is that I don't really think this is the appropriate place for them, as this keeps them scoped to LabKey server and modules defined within it, but it excludes their availability to our packages e.g. labkey-ui-components, labkey-ui-premium, which is where the majority of our react and jest code lives. I also don't really want a solution that involves duplicating these rules across every repo that needs them. That said, I think this is an acceptable place to put them for now, as it allows people to experiment with them today, albeit in a limited scope. We can always move these rules to a more ideal location later.

.agents/review-checklists/jest/code-quality.md
labkey-martyp: I found it really useful to at least have the Assert part with a comment on how it is asserting what is being tested. Some structure and consistency is helpful as well. But I'm not set on AAA, it's just what I found when researching ways to have test clarity for people and AI.

.agents/review-checklists/jest/business-logic.md
labkey-martyp: Ok I added a more specific example.

.agents/review-checklists/jest/business-logic.md
labkey-martyp: I removed it. I think it would be applicable if we were specifically testing our API wrapper framework or something, but probably not doing that anywhere.

.agents/review-checklists/react/code-quality.md
labkey-martyp: Yes it absolutely can. So much easier than doing it manually

.agents/review-checklists/react/business-logic.md
labkey-martyp: Yeah these can be annoying to debug

.agents/review-checklists/jest/performance.md
labkey-martyp: Yeah maybe "redundant" and "near-duplicate" should be split into separate checks. One is performance - not running redundant tests, the other may be more code-quality - sharing common code and using test.each (see example)

.agents/review-checklists/jest/code-quality.md
labkey-martyp: Yeah I think these kinds of things can save us time on code reviews so we can spend more time on higher level thinking.

.agents/review-checklists/jest/code-quality.md
labkey-martyp: Yeah it kind of applies to variables, imports or mock setups. I'm not sure if there's more appropriate terminology?

.claude/skills/code-review-react/skill.md
labkey-martyp: Yeah good call. I think some of this can be combined. I will look into doing that in a follow up PR.

.agents/review-checklists/react/performance.md
labkey-martyp: Can you point me to an example or two in our code? We can get that in a follow up PR.

labkey-martyp:
> I think overall these rules are fine. I did leave some feedback that you should review before merging. If you have follow up questions, or want to have a discussion, let me know.
> 
> My biggest issue with these rules at the moment is that I don't really think this is the appropriate place for them, as this keeps them scoped to LabKey server and modules defined within it, but it excludes their availability to our packages e.g. labkey-ui-components, labkey-ui-premium, which is where the majority of our react and jest code lives. I also don't really want a solution that involves duplicating these rules across every repo that needs them. That said, I think this is an acceptable place to put them for now, as it allows people to experiment with them today, albeit in a limited scope. We can always move these rules to a more ideal location later.

Yep it would be great to discuss a better location. I actually thought everyone was putting labkey-ui-components and labkey-ui-premium under server, but from our meeting I now realize that is not the case.

PR #1295, opened 2026-02-26
bump kotlin for CVE-2020-29582

labkey-willm:
#### Rationale
bump kotlin for CVE-2020-29582

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -189,7 +189,7 @@ httpcore5Version=5.3.6
 httpclientVersion=4.5.14
 httpcoreVersion=4.4.16
 
-intellijKotlinVersion=1.9.10
+intellijKotlinVersion=2.3.10
 
 # Update all Jackson dependency versions below in tandem, unless one gets a patch release out-of-sync with the others
 jacksonVersion=2.20.0


labkey-jeckels:
FWIW, this looks like a bad match from the dependency checker.

The (old) CVE reports that <= 1.4.20 is affected, and 1.4.21 is patched. We're were already on 1.9.10.

Adopting a newer version is likely OK, though it may be flagged as vulnerable.

labkey-willm:
> FWIW, this looks like a bad match from the dependency checker.
> 
> The (old) CVE reports that <= 1.4.20 is affected, and 1.4.21 is patched. We're were already on 1.9.10.
> 
> Adopting a newer version is likely OK, though it may be flagged as vulnerable.

thanks, I saw that, but NVD says up to 2.1.0 is affected, so I thought it best to bump it anyway: https://nvd.nist.gov/vuln/detail/cve-2020-29582

labkey-jeckels:
> > FWIW, this looks like a bad match from the dependency checker.
> > The (old) CVE reports that <= 1.4.20 is affected, and 1.4.21 is patched. We're were already on 1.9.10.
> > Adopting a newer version is likely OK, though it may be flagged as vulnerable.
> 
> thanks, I saw that, but NVD says up to 2.1.0 is affected, so I thought it best to bump it anyway: https://nvd.nist.gov/vuln/detail/cve-2020-29582

Looking at the history in NVD (click to expand the section), there was a modification yesterday to the CPE, which must be why this years-old CVE started firing. There's a link to an equally old Apache mailing list as an alleged source (I think), which seems quite irrelevant.

https://lists.apache.org/thread.html/r2721aba31a8562639c4b937150897e24f78f747cdbda8641c0f659fe%40%3Cusers.kafka.apache.org%3E

Anyway, hopefully the new version is happy in terms of tests and CVE matching. Nice to upgrade to a new version instead of suppressing the bad match.

labkey-adam (approved):

PR #1314, opened 2026-03-20
Compress TSV exports

labkey-jeckels:
#### Rationale
Spring Boot compresses many text HTTP responses by default. But not TSV.

#### Changes
* Add `text/tab-separated-values` to the default set of compressed MIME types

- [x] Code review @labkey-matthewb 
- [x] Manual test - confirm that HTML, JSON, JS, CSV, TTF, SVG, and TSV responses use gzip compression @labkey-tchad

### DIFF
--- a/server/embedded/src/org/labkey/embedded/LabKeyServer.java
+++ b/server/embedded/src/org/labkey/embedded/LabKeyServer.java
@@ -130,9 +130,9 @@ public class LabKeyServer
 
                  // Enable HTTP compression for response content
                  put("server.compression.enabled", "true");
-                 // Spring Boot compresses HTML, JSON and other types by default, but not TSV. We have to duplicate the
-                 // defaults and add text/tab-separated-values
-                 put("server.compression.mime-types", "text/html,text/xml,text/plain,text/css,text/javascript,application/javascript,application/json,application/xml,text/tab-separated-values");
+                 // Spring Boot compresses HTML, JSON and other types by default, but not TSV, CSV, or SVG. We have to
+                 // duplicate the defaults and add those types
+                 put("server.compression.mime-types", "text/html,text/xml,text/plain,text/css,text/javascript,application/javascript,application/json,application/xml,text/tab-separated-values,text/csv,image/svg+xml");
 
                  put("server.tomcat.accesslog.enabled", "true");
                  put("server.tomcat.accesslog.pattern", "%h %l %u %t \"%r\" %s %b %D %S %I \"%{Referer}i\" \"%{User-Agent}i\" %{LABKEY.username}s %{X-Forwarded-For}i");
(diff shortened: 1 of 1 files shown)


labkey-matthewb (approved):

server/embedded/src/org/labkey/embedded/LabKeyServer.java
```diff
@@ -130,6 +130,9 @@ public static void main(String[] args)
 
                  // Enable HTTP compression for response content
                  put("server.compression.enabled", "true");
+                 // Spring Boot compresses HTML, JSON and other types by default, but not TSV. We have to duplicate the
+                 // defaults and add text/tab-separated-values
+                 put("server.compression.mime-types", "text/html,text/xml,text/plain,text/css,text/javascript,application/javascript,application/json,application/xml,text/tab-separated-values");
```
labkey-matthewb: I see one use of text/json in our code, but I don' think we're using that in responses.

labkey-nicka: I do not think `text/json` is a valid mime-type. It should only be `application/json` according to [RFC8259](https://datatracker.ietf.org/doc/html/rfc8259).

labkey-tchad:
TSVs are compressed. What about other common mime types?
- Excel: `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`
- CSV: `text/csv`
- ~SVG: `image/svg+xml`~ (any large svgs are generated in the browser)
- ~PNG: `image/png`~ (already comressed)
- Truetype fonts: `font/ttf` (Roboto and Titillium add up to about 450KB)

labkey-jeckels:
> TSVs are compressed. What about other common mime types?
> 
> * Excel: `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`
> * CSV: `text/csv`
> * ~SVG: `image/svg+xml`~ (any large svgs are generated in the browser)
> * ~PNG: `image/png`~ (already comressed)
> * Truetype fonts: `font/ttf` (Roboto and Titillium add up to about 450KB)

I went with CSV and SVG, as they're text formats that will compress well. TTF is less compelling given their static nature, but I added it.

Excel and PNG are already compressed

labkey-tchad:
CSVs are compressed but none of the SVGs or TTF files are.
Also, no webdav downloads seem to be compressed. Should they be?

labkey-jeckels:
> CSVs are compressed but none of the SVGs or TTF files are. Also, no webdav downloads seem to be compressed. Should they be?

When serving static files, whether they're part of a module or not, the request is served by DavController. It uses Tomcat's `sendfile` "API" to send them, which is more efficient for I/O than Stream-based responses but doesn't support compressing. Therefore, for (most) compressible static files, the build pre-creates a `.gz` version and we serve that instead.

https://github.com/LabKey/platform/blob/7d500be3cfb1119eac1ac77d5bdadc07c7bf6756/core/src/org/labkey/core/webdav/DavController.java#L5004

Dynamically generated content like TSV exports go through a Stream and therefore get compressed per the settings in this change.

I was on the fence about including TTF in the compression list, and just reverted them. They're cached so it shouldn't be much of a difference in real-world perf.

I propose no other changes at this point. The main concern was compressing large TSV exports, which this change accomplishes.

@labkey-matthewb

PR #1319, opened 2026-03-24
Add Claude Skill for checking WCAG compliance

XingY:
#### Related Pull Requests
* 

#### Changes
* Add new wcag-complaince skill to check PR against wcag2.2 standards
* Modify code-review-react skill to additionally check for accessibility


labkey-martyp (approved):
This looks good to me. I don't have as much experience with WCAG but the skill looks good. I tested the standalone and react code reviews on an EHR project. Eventually we'll want to standardize where we're putting the .md files for skills. I have the other code-review-react .md files in the .agents directory but maybe now that we're standardizing on claude I can move them into .claude directory.

PR #1331, opened 2026-04-08
Update hooks path to use $CLAUDE_PROJECT_DIR

labkey-martyp:
#### Rationale
Some claude functionality unnecessarily blocked when claude task navigates to sub-repos.

#### Changes
* pretooluse hooks to use $CLAUDE_PROJECT_DIR for python directory

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "cd \"$(git rev-parse --show-toplevel)\" && if command -v python3 >/dev/null 2>&1; then python3 .claude/hooks/check-dangerous-commands.py; else python .claude/hooks/check-dangerous-commands.py; fi"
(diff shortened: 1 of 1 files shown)


labkey-bpatel (approved):

labkey-jeckels:
Does this need testing on OSX or otherwise? If so, please share a repro steps.

labkey-martyp:
> Does this need testing on OSX or otherwise? If so, please share a repro steps.

@labkey-jeckels I don't really have a repro, you can try to direct claude to look in subrepos for a particular task. Might be good if you could just pull down the change while you're using claude for a quick sanity that the main claude usage is not blocked on mac.

labkey-jeckels:
> > Does this need testing on OSX or otherwise? If so, please share a repro steps.
> 
> @labkey-jeckels I don't really have a repro, you can try to direct claude to look in subrepos for a particular task. Might be good if you could just pull down the change while you're using claude for a quick sanity that the main claude usage is not blocked on mac.

I did some platform PR reviewing and patching with this FB and all worked normally.

PR #1340, opened 2026-04-15
Update to netty version 4.2.12.Final to address CVEs

labkey-susanh:
#### Rationale
We've got some CVE reports coming from an older netty version

#### Changes
* Update to the latest version of netty

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -266,8 +266,8 @@ modelContextProtocolVersion=1.1.1
 
 mssqlJdbcVersion=13.2.1.jre11
 
-# Netty - transitive dependency via azure-core-http-netty; force for CVE-2025-67735
-nettyVersion=4.2.8.Final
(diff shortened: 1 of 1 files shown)


gradle.properties
```diff
@@ -266,8 +266,8 @@ modelContextProtocolVersion=1.1.1
 
 mssqlJdbcVersion=13.2.1.jre11
 
-# Netty - transitive dependency via azure-core-http-netty; force for CVE-2025-67735
-nettyVersion=4.2.8.Final
+# Netty - transitive dependency via azure-core-http-netty; force for CVE-2025-67735, CVE-2026-33871, CVE-2026-33870
```
labkey-jeckels: Do we want to keep a running list of Netty CVEs here? We could just call out the one(s) that prompted the bump to the current version

labkey-jeckels (approved):

PR #1341, opened 2026-04-16
Add Gradle buildscript dependency for LabKey plugins

labkey-tchad:
#### Rationale
IntelliJ doesn't seem to recognize buildscript dependencies that come from `settings.gradle`. This causes class imports from our gradle plugin to show up as errors in the IDE.
With some experimentation, I discovered that adding the dependency to the root `build.gradle` makes IntelliJ recognize the imports throughout the project.
In addition to getting code assistance in our Gradle files, this will prevent IntelliJ's "Optimize Imports" from removing `org.labkey.gradle.util.ExternalDependency` from files that need it.

Before:
image

After:
image

#### Related Pull Requests
* N/A

#### Changes
* Add Gradle buildscript dependency for LabKey plugins

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,10 +1,16 @@
 import org.apache.commons.lang3.SystemUtils
-import org.labkey.gradle.plugin.NpmRun
 import org.labkey.gradle.task.PurgeNpmAlphaVersions
 import org.labkey.gradle.task.PurgeNpmVersions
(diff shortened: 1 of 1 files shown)


labkey-susanh (approved):
Great! I'm glad there's finally a solution for this very annoying behavior.

PR #1353, opened 2026-04-28
Add React guidance for unit tests, JSX structure, and memoization

labkey-jeckels:
#### Rationale
We like unit tests and well structured JSX/TSX. Make sure Claude Code does too.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/7623

#### Changes
* Specific guidance for what we want to see in JSX/TSX
* Specific guidance for unit tests for React code

### DIFF
--- a/.agents/review-checklists/react/code-quality.md
+++ b/.agents/review-checklists/react/code-quality.md
@@ -127,7 +127,6 @@ Flag JSX return statements >25 lines or those with multiple conditional branches
 ### Exceptions / False Positives
 
 - Do not flag simple, single-element returns or small layout wrappers.
-- Do not flag when extraction would require excessive prop-drilling that makes code less readable.
 
 ### Description
 
--- a/.agents/review-checklists/react/code-quality.md
+++ b/.agents/review-checklists/react/code-quality.md
@@ -43,6 +43,184 @@ const MyComponent: FC<MyComponentProps> = ({ report, tab }) => {
 };
 ```
 
+## React components and hooks should have unit tests
+
+**Urgency:** suggestion
+
+### Category
+
+Maintainability
+
+### Confidence Threshold
+
+Flag new or significantly modified components/hooks lacking Jest tests when they contain logic (state, effects, event handlers, conditional rendering). Simple presentational components rendering static JSX may defer testing. Existing untested code not modified in the PR should not be flagged.
+
+### Exceptions / False Positives
+
+- Do not flag demo-only, internal layout wrappers, or placeholder components.
+- Do not flag when tests exist in a different location (e.g., shared test suite) if verifiable.
+- Do not flag existing untested code outside the PR scope.
+
+### Description
+
+Components with state, effects, event handlers, or conditional logic need Jest tests. Test what should always be covered: `useState`/`useReducer`, event handlers, conditional rendering, custom hooks with logic, permission-gated rendering. Layout-only components are optional.
+
+### Anti-patterns to Flag
+
+```tsx
+// ❌ Stateful component with events but no tests
+const Counter: FC = () => {
+    const [count, setCount] = useState(0);
+    return (
+        <div>
+            <p>Count: {count}</p>
+            <button onClick={() => setCount(count + 1)}>Increment</button>
+        </div>
+    );
+};
+
+// ❌ Custom hook with async logic but no tests
+const useCustomFetch = (url: string) => {
+    const [data, setData] = useState(null);
+    useEffect(() => {
+        fetch(url).then(res => res.json()).then(setData);
+    }, [url]);
+    return data;
+};
+```
+
+### Suggested Fix
+
+Create `Component.test.tsx` with meaningful assertions per [`jest/business-logic.md`](../../jest/business-logic.md):
+
+```tsx
+test('increments count when button clicked', async () => {
+    render(<Counter />);
+    await userEvent.click(screen.getByRole('button', { name: /increment/i }));
+    expect(screen.getByText('Count: 1')).toBeInTheDocument();
+});
+```
+
+### How to Detect
+
+1. Check for `.test.tsx`/`.test.ts` file corresponding to modified component.
+2. If missing, flag if component has: hooks, event handlers, or conditional logic.
+3. If tests exist, verify they assert on behavior (not just render existence).
+
+## Large JSX/TSX blocks should be extracted into separate components
+
+**Urgency:** suggestion
+
+### Category
+
+Maintainability
+
+### Confidence Threshold
+
+Flag JSX return statements >50 lines or those with multiple conditional branches (3+ `{condition && <...>}` or ternary operators). Simple single-element returns or small layout wrappers are exempt.
+
+### Exceptions / False Positives
+
+- Do not flag simple, single-element returns or small layout wrappers.
+- Do not flag when extraction would require excessive prop-drilling that makes code less readable.
+
+### Description
+
+Large JSX blocks reduce readability, increase cognitive load, and make testing harder. Red flags: multiple conditional rendering branches, repeated patterns (like tab buttons with similar structure), deeply nested element hierarchies, and components managing multiple independent concerns.
+
+### Anti-patterns to Flag
+
+```tsx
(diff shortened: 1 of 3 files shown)


.agents/review-checklists/react/code-quality.md
```diff
@@ -43,6 +43,184 @@ const MyComponent: FC<MyComponentProps> = ({ report, tab }) => {
...
+    useEffect(() => {
+        fetch(url).then(res => res.json()).then(setData);
+    }, [url]);
+    return data;
+};
+```
+
+### Suggested Fix
+
+Create `Component.test.tsx` with meaningful assertions per [`jest/business-logic.md`](../../jest/business-logic.md):
```
labkey-martyp: Shouldn't this path just be ../jest/business-logic.md?

.agents/review-checklists/react/code-quality.md
```diff
@@ -43,6 +43,184 @@ const MyComponent: FC<MyComponentProps> = ({ report, tab }) => {
...
+};
+
+// ❌ Custom hook with async logic but no tests
+const useCustomFetch = (url: string) => {
+    const [data, setData] = useState(null);
+    useEffect(() => {
+        fetch(url).then(res => res.json()).then(setData);
+    }, [url]);
+    return data;
+};
```
labkey-martyp, line 94: Claude is flagging this as bad code example because it's missing error handling and cleanup, which could encourage an LLM to use this. Here's the recommended code.
```
// ❌ Custom hook with async logic but no tests
  const useCustomFetch = <T,>(url: string): T | null => {
      const [data, setData] = useState<T | null>(null);
      useEffect(() => {
          const controller = new AbortController();
          fetch(url, { signal: controller.signal })
              .then(res => res.json())
              .then(setData)
              .catch(() => { /* ignore aborts/errors for brevity */ });
          return () => controller.abort();
      }, [url]);
      return data;
  };
```

labkey-martyp (approved):
I haven't tried it out, but looks like a good rule.

.agents/review-checklists/react/code-quality.md
labkey-jeckels: This is interesting. It IS a bad code example, though its intent is to be bad in a different way (lacking tests, not bad error handling).

Do we need our examples of bad code to be good in ways other than the specific problem they're intending to call out? The downside is overly verbose examples and token bloat.

.agents/review-checklists/react/code-quality.md
labkey-jeckels: Yes. Fixed.

.agents/review-checklists/react/code-quality.md
labkey-alan: I am not sure how useful these code examples are because we cannot use fetch anywhere, and as a result cannot use AbortControllers. We'd need to release a 2.x version of `@labkey/api` for this to be relevant to us. This is something we should do, and I was just talking with Nick about it the other day because I want to use AbortControllers, but there isn't really a path forward for us to use either because all of our API requests are done via `Ajax.request`

labkey-martyp: This was a low priority call out by claude, but I suspect what you might get is an LLM adds tests to the bad example and thinks it is now good code to use as a reference. Even though it has other issues. Maybe something to research further.

labkey-jeckels: Updated to `async function` and a comment to say things were omitted for brevity. Does that seem reasonable to you both?

labkey-jeckels:
I pushed some more updates based on feedback from the plate designer migration.

My understanding is that `CLAUDE.md` is the right place to put guidance to help Claude generate new code according to our conventions. I added notes based on things it did repeatedly that I refactored prior to merging.

The guidance in the other files I think is more general purpose for all reviews, regardless of who wrote the code.

I'll re-request review based on these updates.

labkey-martyp (approved):
Just the one merge issue, otherwise looks good.

labkey-jeckels:
> Just the one merge issue, otherwise looks good.

Fixed. I'll wait for @labkey-alan to have a chance to review too.

.agents/review-checklists/react/code-quality.md
```diff
@@ -43,6 +43,269 @@ const MyComponent: FC<MyComponentProps> = ({ report, tab }) => {
...
+Maintainability
+
+### Confidence Threshold
+
+Flag JSX return statements >25 lines or those with multiple conditional branches (3+ `{condition && <...>}` or ternary operators). Simple single-element returns or small layout wrappers are exempt.
+
+### Exceptions / False Positives
+
+- Do not flag simple, single-element returns or small layout wrappers.
+- Do not flag when extraction would require excessive prop-drilling that makes code less readable.
```
labkey-alan: I disagree with this line and think we should remove it. It's incredibly rare that a single large block of JSX being split into smaller chunks would result in excessive prop-drilling. For that you'd need a wildly complicated block of JSX that needs to be split into multiple nested components, but even in that case prop drilling is preferred to a huge unreadable, hard to test, block of JSX. Also, in those cases you can leverage things like react context to avoid prop drilling.

I also think this wording is too vague to be enforceable, what exactly qualifies as excessive? Without a clear definition of that I don't think we'd get consistent results, and I think it could easily be justified as a reason to ignore otherwise sound feedback.

If we find that not having this exception leads to prop-drilling issues in practice, we can revisit with a concrete example. For now I think just removing it is the right call.

.agents/review-checklists/react/performance.md
```diff
@@ -2,6 +2,56 @@
...
+const getHandler = useCallback((id: string) => () => handleClick(id), [handleClick]);
+<Item onClick={getHandler(item.id)} />
+```
+
+### Suggested Fix
+
+```tsx
+const handleDelete = useCallback(() => {
+    doDelete(item.id);
+}, [item.id]);
```
labkey-alan:
```suggestion
}, [doDelete, item.id]);
```

CLAUDE.md
```diff
@@ -83,6 +83,13 @@ We keep local copies of the `@labkey` packages in the `clientAPIs/` directory. Y
 
 The local copies of the packages may contain changes related to the current branches in any of the modules that have an NPM build. For example there may be changes to the `@labkey/components` package that affect the package in the `server/modules/platform/core` module. These packages are not required to be present to build the project, so they may not be available. If they are not present, you can assume that there are no changes in those packages relevant to the current branch. If you suspect an issue is with one of the packages, but it is not present you may prompt the user to check out a local copy of the relevant package.
 
+### React Conventions
+- We are using React 18
+- **Component typing:** Use `React.FC<Props>` (or `React.FC` for no-prop components) for proper components used as `<Component />`. Use `: React.ReactNode` as the return type for render helpers — functions that return JSX but are not used as components (typically named with a lowercase letter or a `render` prefix). Do **not** annotate with `: JSX.Element` or `: React.ReactElement`.
```
labkey-alan:
```suggestion
- **Component typing:** Use `FC<Props>` (or `FC` for no-prop components) for proper components used as `<Component />`. Use `ReactNode` as the return type for render helpers — functions that return JSX but are not used as components (typically named with a lowercase letter or a `render` prefix). Do **not** annotate with `JSX.Element` or `ReactElement`.
- **React type imports:** Always import React types directly from `react` rather than accessing them via the `React.*` namespace. For example: `import { FC, ReactNode } from 'react'`, never `React.FC` or `React.ReactNode`.
```

labkey-alan (approved):
I think overall this looks good, but I did leave some feedback that I think should be included. I'm approving so you can easily make the changes and merge, but if you disagree with my feedback let's have a discussion before merging.

.agents/review-checklists/react/code-quality.md
labkey-jeckels: Removed.

labkey-jeckels:
> I think overall this looks good, but I did leave some feedback that I think should be included. I'm approving so you can easily make the changes and merge, but if you disagree with my feedback let's have a discussion before merging.

Thanks. I accepted your suggestions.

PR #1368, opened 2026-05-08
Remove versions for GWT and Jakarta annotations

labkey-jeckels:
#### Rationale
We no longer depend on GWT.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/7659

#### Changes
* Remove version numbers for now irrelevant dependencies

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,7 @@ allprojects {
                 analyzers.ossIndex.enabled = false
             }
             formats = ['HTML', 'JUNIT']
-            skipConfigurations = ['dedupe', 'gwtCompileClasspath', 'gwtRuntimeClasspath', 'developmentOnly']
+            skipConfigurations = ['dedupe', 'developmentOnly']
             skipProjects = [':server:testAutomation']
 
             nvd {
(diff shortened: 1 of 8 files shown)


labkey-jeckels:
@labkey-susanh My local build still works without these properties. Let me know if if it's best to wait for the Gradle plugin update before merging this. https://github.com/LabKey/internal-issues/issues/1143

build.gradle
```diff
@@ -48,7 +48,7 @@ allprojects {
                 analyzers.ossIndex.enabled = false
             }
             formats = ['HTML', 'JUNIT']
-            skipConfigurations = ['dedupe', 'gwtCompileClasspath', 'gwtRuntimeClasspath', 'developmentOnly']
+            skipConfigurations = ['dedupe', 'developmentOnly']
```
labkey-jeckels, line 51: @labkey-susanh let me know if this should remain until the Gradle plugin updates.

labkey-susanh: If there are no modules with a directory called `gwtsrc` then there should be no modules that apply the GWT plugin, and thus none that have these configurations, so I believe it's safe to remove them. There's also, I think, no harm in leaving them and just making a note in the issue that is already open to clean these up when the plugins are updated to remove the GWT plugin.

labkey-susanh (approved):

PR #1379, opened 2026-05-18
Suppress Kiota CVEs

labkey-adam:
#### Rationale
OWASP dependency checker seems to be confusing our kiota libraries (https://github.com/microsoft/kiota-java) with the kiota tool (https://github.com/microsoft/kiota)


labkey-adam:
Merged to 26.5 instead

PR #1383, opened 2026-05-20
Address Azure-related findings

labkey-adam:
#### Rationale
OWASP dependency checker seems to be confusing our kiota libraries (https://github.com/microsoft/kiota-java) with the kiota tool (https://github.com/microsoft/kiota)

### DIFF
--- a/dependencyCheckSuppression.xml
+++ b/dependencyCheckSuppression.xml
@@ -385,4 +385,15 @@
         <packageUrl regex="true">^pkg:maven/com\.microsoft\.kiota/microsoft-kiota-serialization-text@.*$</packageUrl>
         <cve>CVE-2026-41134</cve>
     </suppress>
+
+    <!--
+    Checker is confusing json-schema-validator with Nu Html Checker
+    -->
+    <suppress>
+        <notes><![CDATA[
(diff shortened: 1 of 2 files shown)


dependencyCheckSuppression.xml
```diff
@@ -319,17 +319,122 @@
...
+    <!--
+    Checker is confusing json-schema-validator with Nu Html Checker
+    -->
+    <suppress>
+        <notes><![CDATA[
+    file name: json-schema-validator-3.0.1.jar
+    ]]></notes>
+        <packageUrl regex="true">^pkg:maven/com\.networknt/json-schema-validator@.*$</packageUrl>
+        <cve>CVE-2025-15104</cve>
+    </suppress>
```
labkey-tchad, line 398: I don't recognize this one.

labkey-tchad (approved):
The dependency checker actually keeps a list of false-positives. We're good to suppress here but I'll open some issues to get them added to the official list.

dependencyCheckSuppression.xml
labkey-adam: This snuck in very recently. May have been pulled in by the latest Azure Identity version, not sure

PR #1356, opened 2026-04-29
Enable IntelliJ MCP, TeamCity guidance, and review-lk

labkey-jeckels:
#### Tasks
- [x] Claude code review
- [x] Test `/review-lk` on OSX @labkey-tchad 
- [x] Test `/review-lk` on Windows @labkey-adam 
- [x] Test `/branch-status`, including `--monitor` (and if brave `--fix`) @labkey-tchad 

---

#### Rationale
IntelliJ has native support for MCP, compatible with Claude Code. Plugins provide other tools like debugger integration.

We can also consolidate the `/review-pr` and `/review-local` commands into a single `/review-lk` that uses a script to gather changes across all matching feature branches (point at a PR or FB), or looks at local changes.

https://www.jetbrains.com/help/idea/mcp-server.html

https://plugins.jetbrains.com/plugin/29233-debugger-mcp-server

https://plugins.jetbrains.com/plugin/29174-ide-index-mcp-server

We can also teach Claude to review TeamCity results and help analyze them. `/branch-status` command will look for branches where have you recent commits, or takes a branch name as an argument. It summarizes pull request and test results. `--monitor` keeps checking for new results and sends a push notification on meaningful updates. `--fix` tells Claude to try its best to fix new failures.

#### Changes
* Tell Claude to prefer the IntelliJ tools instead of find/grep
* Skills provided by third-party plugins too
* `/review-lk` command and supporting Python script
* Guidance for TeamCity MCP
* `/branch-status` command to assess PR and TeamCity results on a branch
* `upgrade-test-author` skill to help create schema upgrade tests

PR #1385, opened 2026-05-21
Add 'tc-branch-cleanup' skill for canceling TeamCity builds

labkey-tchad:
#### Rationale
After merging branches, there are often builds left queued on TeamCity. It would be handy to be able to clear out those builds so they don't consume TeamCity resources.
`tc_branch_builds.py` lists queued and running builds for a branch.
The `tc-branch-cleanup` leverages that script to cancel builds from a particular branch.

The script also works from the command line, without claude.
```
python3 .claude/scripts/tc_branch_builds.py my-branch --ids \
    | xargs -I{} teamcity run cancel {} --yes --comment "<comment>"
```

#### Related Pull Requests
* N/A

#### Changes
* Add script to list queued and running builds for a branch
* Add 'tc-branch-cleanup' skill for canceling TeamCity builds

### DIFF
--- a/.claude/scripts/tc_branch_builds.py
+++ b/.claude/scripts/tc_branch_builds.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python3
+"""List queued and running TeamCity builds for a given branch."""
+
+import argparse
+import json
+import subprocess
+import sys
+
+
+def tc_api_all(endpoint: str) -> list[dict]:
+    """Fetch all pages from a REST endpoint using --paginate --slurp."""
+    result = subprocess.run(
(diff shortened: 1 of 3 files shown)


labkey-jeckels (approved):
I used this to kill some builds that weren't needed. It worked but I had a couple of issues:

1. It requires install of the TeamCity CLI. I didn't have it. I updated the guidance for getting it installed.
2. It didn't find the feature branch because TeamCity doesn't use the `fb_` prefix. I think this should be handled automatically when combined with the guidance in https://github.com/LabKey/server/pull/1356

PR #1392, opened 2026-05-26
Add some instructions for Claude regarding accessibility reviews

labkey-susanh:
#### Rationale
Having gone through a round of fixing up various accessibility issues in our apps. we want Claude to review new functionality for accessibility impact, so we add some general instructions in the review commands. I've not tried this on a branch that wasn't specifically for addressing accessibility issues.

@labkey-jeckels This may be made obsolete by the changes in your other branch, in which case I can close this. 

#### Related Pull Requests
* 

#### Changes
* Update review commands with an Accessibility phase.
* Update wcag checklist with specifics about callbacks.

### DIFF
--- a/.claude/commands/review-local.md
+++ b/.claude/commands/review-local.md
@@ -56,6 +56,11 @@ Compare what the code *actually does* (from Phase 2) against what it *should do*
 - Are there other places in the codebase that should have been updated alongside this change?
 - Are tests updated to cover the new behavior?
 
+## Phase 6: Accessibility
+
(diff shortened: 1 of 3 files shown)


cnathe (approved):

cnathe (commented):
Looks like Alan has a PR open to move these to labkey-workspaces, so it might be easier to hold off and reapply these small changes there after they are moved.

labkey-susanh:
Closing for now. We can add accessibility instructions after these files have moved, if we find them lacking.

PR #1398, opened 2026-05-29
Latest Spring AI milestone release

labkey-adam:
#### Rationale
Getting closer to GA

Pulls in `mcp-2.0.0-M8`, which does not fix the issue we reported


labkey-nicka (approved):
Expression assistant works against this version (ported ahead to develop).

PR #1467, opened 2026-08-07
XXE warning for JAXP

labkey-jeckels:
## Rationale
We want to avoid introducing new XXE problems in the future.

## Related Pull Requests
- https://github.com/LabKey/platform/pull/7875

## Changes
- IntelliJ warning to prevent use for future non-hardened SchemaFactories

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -1,6 +1,5 @@
 <component name="InspectionProjectProfileManager">
   <profile version="1.0">
-    <option name="myName" value="Project Default" />
     <inspection_tool class="35f0afe5-6749-3bcd-88fe-638b65ec3b43" enabled="true" level="ERROR" enabled_by_default="true" editorAttributes="ERRORS_ATTRIBUTES" />
     <inspection_tool class="4343478d-72f3-35b1-a08b-33b2219c1ba0" enabled="true" level="ERROR" enabled_by_default="true" editorAttributes="ERRORS_ATTRIBUTES" />
     <inspection_tool class="9c4146d5-e6f9-3c25-a503-0c0b81240a75" enabled="true" level="ERROR" enabled_by_default="true" editorAttributes="ERRORS_ATTRIBUTES" />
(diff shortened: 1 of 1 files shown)


.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -165,6 +164,10 @@
       <searchConfiguration name="XMLInputFactory.newInstance()" text="javax.xml.parsers.DocumentBuilderFactory.newInstance()" recursive="true" caseInsensitive="true" type="JAVA" pattern_context="default">
         <constraint name="__context__" within="" contains="" />
       </searchConfiguration>
+      <searchConfiguration name="SchemaFactory.newInstance()" description="XML parsers should always be configured to protect against XXE — use XmlBeansUtil.getHardenedSchemaFactory()" suppressId="SchemaFactory" text="javax.xml.validation.SchemaFactory.newInstance($schemaLanguage$)" recursive="true" caseInsensitive="true" type="JAVA" pattern_context="default">
```
labkey-bpatel: Looks like the description should say `use XmlBeansUtil.schemaFactory()` instead of `use XmlBeansUtil.getHardenedSchemaFactory()`. I don't see a `getHardenedSchemaFactory()` method in XmlBeansUtil.

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -1,6 +1,5 @@
 <component name="InspectionProjectProfileManager">
   <profile version="1.0">
-    <option name="myName" value="Project Default" />
```
labkey-bpatel, line 3: Was this removal intentional?

labkey-bpatel (approved):
Approving, but please take a look at my comments.

.idea/inspectionProfiles/Project_Default.xml
labkey-jeckels: It seemed harmless (I don't know what it does) but I restored it, just in case.

.idea/inspectionProfiles/Project_Default.xml
labkey-jeckels: Updated to match the final method name.

